PR #8, opened 2019-06-25
Iiif 281 update open api spec

ksclarke:
* Update the OpenAPI spec to match patterns described in IIIF-270
* Updated tests to match the new endpoint URLs


ksclarke:
Travis seems to be having some troubles reaching the Maven central repo so that reminded me to add a local Maven cache for the Travis build too. Added in new commit pushed to this branch.

PR #15, opened 2019-07-19
[IIIF-248] Set up the Travis to Travis link for DockerHub push of manifest-store

hardyoyo:


.travis/trigger.sh
```diff
@@ -0,0 +1,140 @@
...
+
+}
+
+ensure_repo_set() {
+  if [[ -z "$repo" ]]; then
+    echo "Error:
+    A repository was expected to be supplied as the
+    argument but an empty string was received.
+
+    Usage:
```
ksclarke: Update usage to reflect the script name change to trigger.sh

.travis/trigger.sh
```diff
@@ -0,0 +1,140 @@
...
+  if [[ "$request_status_code" =~ 20[02] ]]; then
+    echo "
+    Success! Build for repository $repo triggered.
+    "
+    exit 0
+   elif [[ "$request_status_code" == "500" ]]; then
+    echo "
+    Build for repository $repo requested, internal service error 500 received, this usually means the request was malformed. Debug time!
+    "
+    exit 0
```
ksclarke: Wouldn't we want to fail the travis build on a 500 response?

PR #20, opened 2019-09-16
[IIIF-444] Use temporary JsonObjects to compare expectedManifest and foundManifest

hardyoyo:


ksclarke:
There are more JSON string compares in the project that should be done at the same time as this if we're going to do this, I think.

hardyoyo:
IIIF-444 has been revised to reflect our desire for a larger refactor. Closing this PR as incomplete.

PR #22, opened 2019-09-24
[IIIF-319] Improve docs for Maniftest-Store's API

hardyoyo:
* Port CSS style from Bucketeer
* Add examples to OpenAPI schema for Manifest-Store
* Add links to manifest documentation


hardyoyo:
I completely missed that the Bucketeer styles changed with that PR, I'll port the diffs over to this PR

ksclarke:
This yellow on white example is hard to read. This is something for which there isn't a corresponding Bucketeer CSS change. I didn't realize you could provide examples with the requests too. That would be a good thing to add to the Bucketeer spec.

_manifeststore_yellow_text

PR #23, opened 2019-09-27
[IIIF-285] Add Locust load test, ignore IDEA-related files

hardyoyo:


hardyoyo:
tested locally against a dev instance of manifest-store, this test works

PR #31, opened 2019-11-21
[IIIF-504] Add collection manifest generation

ksclarke:
* Update Snyk badge on README to come directly from snyk.io
* Update vertx version and adjust tests accordingly
* Add abstract handler and verticle
* Add CSV parsing
* Add collection manifest builder
* Remove getLogger() from abstract derived classes for better logging output


src/main/java/edu/ucla/library/iiif/fester/CSV.java
```diff
@@ -0,0 +1,41 @@
+
+package edu.ucla.library.iiif.fester;
+
+/**
+ * CSV headers that we can about for our manifest generation.
```
hardyoyo: typo? "care?"

ksclarke: Yes, thanks. Will fix.

src/main/java/edu/ucla/library/iiif/fester/verticles/MainVerticle.java
```diff
@@ -111,6 +116,51 @@ public void start(final Future<Void> aFuture) {
         });
     }
 
+    // Start verticles -- this is where to add any new verticles that we create and want to load
```
hardyoyo: :tada: thanks!

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
```diff
@@ -0,0 +1,216 @@
...
+                    final List<Manifest> manifests = works.get(collectionID);
+
+                    // If we have work manifests, add them to the collection manifest
+                    if (manifests != null) {
+                        manifestList.addAll(manifests);
+                    } else {
+                        LOGGER.warn(MessageCodes.MFS_118, collectionID);
+                    }
+
+                    // We're going to send to the S3 verticle
```
hardyoyo: a TODO for IIIF-507 right here would be handy, but not required

PR #33, opened 2019-12-02
[IIIF-507] Borrow the S3BucketVerticle from Bucketeer, adjust for Fester

hardyoyo:
- Remove the ResultsMap and other image-related things from the
Bucketeer S3BucketVerticle
- Refactor the message requirements for the S3BucketVerticle to work
with a JSON object
- Refactor the test for the S3BucketVerticle to send a manifest instead
of an image
- Bump the version of Vert-X we're using

### DIFF
--- a/src/test/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticleTest.java
+++ b/src/test/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticleTest.java
@@ -1,0 +1,189 @@
+
+package edu.ucla.library.iiif.fester.verticles;
+
+import static edu.ucla.library.iiif.fester.Constants.MESSAGES;
+import static org.junit.Assume.assumeTrue;
+
+import java.util.UUID;
+
+import org.junit.After;
+import org.junit.AssumptionViolatedException;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3Client;
+
+import info.freelibrary.util.Logger;
+import info.freelibrary.util.LoggerFactory;
+
+import edu.ucla.library.iiif.fester.Config;
+import edu.ucla.library.iiif.fester.Constants;
+import edu.ucla.library.iiif.fester.MessageCodes;
+import io.vertx.config.ConfigRetriever;
+import io.vertx.core.DeploymentOptions;
+import io.vertx.core.Vertx;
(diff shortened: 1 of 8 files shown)


src/test/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticleTest.java
```diff
@@ -0,0 +1,189 @@
...
+                        myAmazonS3 = new AmazonS3Client(myAWSCredentials);
+                    }
+
+                    s3Bucket = config.getString(Config.S3_BUCKET);
+                    LOGGER.debug(MessageCodes.MFS_067, getClass().getName());
+                    asyncTask.complete();
+                });
+            } else {
+                aContext.fail(getConfig.cause());
+                asyncTask.complete();
```
ksclarke: Don't need to complete the asyncTask after an explicit fail. i know this was copied from the bucketeer class but we probably clean it up there too. It generates a warning because the context has already been closed by the time the complete is called.

src/test/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticleTest.java
```diff
@@ -0,0 +1,189 @@
...
+                }
+
+                vertx.deployVerticle(VERTICLE_NAME, options.setConfig(config), deployment -> {
+                    if (deployment.failed()) {
+                        final Throwable details = deployment.cause();
+                        final String message = details.getMessage();
+
+                        LOGGER.error(details, message);
+                        aContext.fail(message);
+                    }
```
ksclarke: Know this is how it is in bucketeer, but it would be good to put the non-`if` stuff here in an `else` since we don't need it to run if the deployment has failed.

src/test/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticleTest.java
```diff
@@ -0,0 +1,189 @@
...
+     *
+     * @param aContext A test context
+     * @throws Exception If there is trouble closing down the Vert.x instance
+     */
+    @After
+    public void tearDown(final TestContext aContext) throws Exception {
+        final Async async = aContext.async();
+
+        myRunTestOnContextRule.vertx().close(result -> {
+            if (!result.succeeded()) {
```
ksclarke: Could just be `if (result.failed())` for clarity?

src/test/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticleTest.java
```diff
@@ -0,0 +1,189 @@
...
+                final String message = LOGGER.getMessage(MessageCodes.MFS_066);
+
+                LOGGER.error(message);
+                aContext.fail(message);
+            }
```
ksclarke: Would be good to put what's not intended to be run after a failure in an `else`

PR #34, opened 2019-12-02
[IIIF-570] Fix error in fester_messsages.xml, add xml-maven-plugin to validate XML

hardyoyo:


hardyoyo:
NOTE: the plugin added by this PR does not catch transposition errors of the type also fixed in this PR. It *does* catch very malformed XML, though, so it'll be helpful.

pom.xml
```diff
@@ -638,6 +638,27 @@
           </execution>
         </executions>
       </plugin>
+      
+      <!--  Validate all XML and XSL in the project during the validate phase -->
+      <plugin>
+        <groupId>org.openjax.xml</groupId>
+        <artifactId>xml-maven-plugin</artifactId>
+        <version>0.9.2</version>
```
ksclarke: Put the version in a variable (e.g. `xml.plugin.version`) and put the variable with the value up in the properties file of the POM file. Consult how the other plugins in the POM are done for an example.

PR #36, opened 2019-12-06
[IIIF-507] Borrow the S3BucketVerticle from Bucketeer, adjust for Fester

hardyoyo:
- Remove the ResultsMap and other image-related things from the
Bucketeer S3BucketVerticle
- Refactor the message requirements for the S3BucketVerticle to work
with a JSON object
- Refactor the test for the S3BucketVerticle to send a manifest instead
of an image
- Bump the version of Vert-X we're using

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticle.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticle.java
@@ -1,0 +1,264 @@
+
+package edu.ucla.library.iiif.fester.verticles;
+
+import static edu.ucla.library.iiif.fester.Constants.EMPTY;
+
+import com.amazonaws.regions.RegionUtils;
+
+import info.freelibrary.util.Logger;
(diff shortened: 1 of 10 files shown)


src/main/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticle.java
```diff
@@ -0,0 +1,264 @@
...
+    @Override
+    @SuppressWarnings({ "deprecation" })
+    public void start(final Future<Void> aFuture) {
+        getJsonConsumer().handler(message -> {
+
+            // grab the Fester configuration
+            final JsonObject config = config();
+
+            // Initialize the S3BucketVerticle by getting our s3 configs and setting up the S3 client
+            if (myS3Client == null) {
```
ksclarke: Move this and the `config()` stuff outside of the `getJsonConsumer().handler(...)`. We want this inside `start()` so it's done when the verticle is started, but don't need to make the check on every message we receive.

PR #40, opened 2019-12-13
[IIIF-574] Add Festerize script for uploading CSVs to Fester

hardyoyo:


src/main/scripts/festerize/README.md
```diff
@@ -0,0 +1,58 @@
...
+                   https://iiif.library.ucla.edu
+  --endpoint TEXT  Service endpoint to use. Default: /collections
+  --out TEXT       Folder to store the results of festerizing. Default:
+                   ./output
+  --loglevel TEXT  Log level for Festerizer logs. Default: INFO, can also be
+                   DEBUG or ERROR
+  --help           Show this message and exit.
+```
+
+Note that the SRC argument above supports standard [filename globbing](https://en.wikipedia.org/wiki/Glob_(programming)
```
ksclarke: Think you're missing a closing paren here? The output gets mangled: https://github.com/UCLALibrary/fester/tree/IIIF-574-csv-uploading/src/main/scripts/festerize

src/main/scripts/festerize/README.md
```diff
@@ -0,0 +1,58 @@
+# festerize
+Script to upload CSV files to the UCLA Library Fester service.
+
+# Using festerize.py
```
ksclarke: I'd say use `##` or `###` here to make it smaller than the page's header.

PR #43, opened 2019-12-17
[IIIF-592] Pass iiifhost via a command line option, fix two path issues

hardyoyo:


hardyoyo:
not ready yet, needs docs

PR #47, opened 2019-12-18
[IIIF 509] Add manifest URLs to the batch CSV output

ksclarke:
* Update the CSV response to that it contains URLs to manifests and collection documents.
* Update tests to expect the new CSVs that come out of PostCsvHandler
* Add tests for for the new LinkUtils utility class.


src/main/java/edu/ucla/library/iiif/fester/handlers/PostCsvHandler.java
```diff
@@ -91,27 +103,46 @@ public void handle(final RoutingContext aContext) {
...
             });
         }
     }
 
+    private void updateCSV(final String aFileName, final String aFilePath, final String aHost,
+            final HttpServerResponse aResponse) {
+        final String responseMessage = LOGGER.getMessage(MessageCodes.MFS_038, aFileName, aFilePath);
+        final FileSystem fileSystem = myVertx.fileSystem();
+
+        // For a first pass, we just return the same CSV that was sent to us.
```
hardyoyo: is this comment still correct?

src/test/java/edu/ucla/library/iiif/fester/handlers/PostCsvHandlerTest.java
```diff
@@ -250,4 +260,32 @@ public final void testWorksCSV(final TestContext aContext) {
             }
         });
     }
+
+    /**
+     * Checks our CSV structures.
```
hardyoyo: this is a very cool method, I like it a lot!

src/main/java/edu/ucla/library/iiif/fester/handlers/PostCsvHandler.java
ksclarke: Nope, thanks! I'll remove it.

PR #51, opened 2020-01-06
[IIIF-618] Fix timeout issue on image w/h lookup

ksclarke:
* Fix the production log location (doesn't affect log location when running in test mode)
* Increase the read timeout when looking for w/h from info.json in Cantaloupe
* Ignore the output directory festerize creates if it's run from the project's root directory
* Make the fallback w/h if image can't be found 1 (a workaround for issue with presentation library)

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/ImageInfoLookup.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/ImageInfoLookup.java
@@ -26,6 +26,8 @@ public class ImageInfoLookup {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ImageInfoLookup.class, Constants.MESSAGES);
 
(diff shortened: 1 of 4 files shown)


src/main/java/edu/ucla/library/iiif/fester/ImageInfoLookup.java
```diff
@@ -61,8 +68,8 @@ public ImageInfoLookup(final String aURL) throws MalformedURLException, IOExcept
                     jsonObject = new JsonObject(result.toString());
 
                     // Find our image's width and height or use zero if they're missing in the manifest
```
markmatney: ~zero~ one

ksclarke: Thanks for catching. Pushed update.

PR #52, opened 2020-01-07
[IIIF-618] Add log warning and one more default value workaround

ksclarke:
* Log a warning when we set the w/h to a default value; this isn't an error but we shouldn't blindly do it
* Add the w/h "set to 1 instead of 0" workaround to one other place that's another on-error fallback

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/ImageInfoLookup.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/ImageInfoLookup.java
@@ -68,10 +68,10 @@ public class ImageInfoLookup {
                     jsonObject = new JsonObject(result.toString());
 
                     // Find our image's width and height or use one if they're missing in the manifest
-                    myHeight = jsonObject.getInteger("height", 1);
-                    myWidth = jsonObject.getInteger("width", 1);
+                    myHeight = jsonObject.getInteger("height", 0);
+                    myWidth = jsonObject.getInteger("width", 0);
 
-                    if (myHeight == 1 || myWidth == 1) {
+                    if (myHeight == 0 || myWidth == 0) {
                         LOGGER.warn(MessageCodes.MFS_073, aURL);
                     }
                 }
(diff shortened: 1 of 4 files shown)


src/main/java/edu/ucla/library/iiif/fester/ImageInfoLookup.java
```diff
@@ -68,8 +68,12 @@ public ImageInfoLookup(final String aURL) throws MalformedURLException, IOExcept
                     jsonObject = new JsonObject(result.toString());
 
                     // Find our image's width and height or use one if they're missing in the manifest
```
markmatney: zero

src/main/resources/fester_messages.xml
```diff
@@ -86,6 +86,7 @@
   <entry key="MFS-070">Unable to retrieve '{}' from the IIIF image server</entry>
   <entry key="MFS-071">Unexpected response code from the IIIF image server: {} [{}]</entry>
   <entry key="MFS-072">Looking up image width and height from: {}</entry>
+  <entry key="MFS-073">Setting image w/h to a questionable default value: 1 [item: {}]</entry>
```
markmatney: 0

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
```diff
@@ -507,13 +507,15 @@ private void addPages(final CsvHeaders aCsvHeaders, final List<String[]> aPageLi
                     // If we've not processed any, check to sequence to find one
                     final List<Canvas> canvases = aSequence.getCanvases();
 
-                    // If there is one use that; else, just use zeros for the w/h values
+                    // If there is one use that; else, just use ones for the w/h values
```
markmatney: zeroes

src/main/resources/fester_messages.xml
markmatney: I think it would be good to say _why_ we're setting the w/h to the questionable default (error(s) in the info.json).

Also, I'm not sure that it's necessarily true that both values would be set to `0` (which the error message could be construed to imply); is it possible to have generated an info.json that omits `width` but not `height`, or vice versa? If so, then the code in `ImageInfoLookup` could set one of those values to `0` and the other to something non-zero.

In either case, we have a non-comformant info.json, so IMO the error message should really be about that.

ksclarke: Good points. I've updated the message text to better reflect what's happening.

markmatney: Looks good

PR #53, opened 2020-01-16
[IIIF-631] Add support for viewingHint and viewingDirection

ksclarke:
* Add support for viewingHint and viewingDirection
* Fix CsvHeaders default values for index positions
* Make item sequence methods in CsvHeaders conform to existing naming pattern
* Add logging for registering the verticle ID map in shared data
* Fix FakeS3BucketVerticle so that it records its deployment ID
* Add tests and test resources for viewingHint/viewingDirection


src/test/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticleTest.java
```diff
@@ -91,6 +119,39 @@ public void tearDown(final TestContext aContext) {
...
+        message.put(Constants.FESTER_HOST, MANIFEST_HOST);
+
+        LOGGER.debug(MessageCodes.MFS_120, SINAI, ManifestVerticle.class.getName());
+
+        myVertx.eventBus().request(ManifestVerticle.class.getName(), message, request -> {
+            if (request.succeeded()) {
+                final JsonObject manifest = new JsonObject(myVertx.fileSystem().readFileBlocking(jsonFile));
+
+                assertEquals(ViewingHint.Option.PAGED.toString(), manifest.getString("viewingHint"));
+                assertEquals(ViewingDirection.RIGHT_TO_LEFT.toString(), manifest.getString("viewingDirection"));
```
markmatney: Should these string literals be sourced from the jiiify-presentation constants, or would that be too verbose?

src/test/java/edu/ucla/library/iiif/fester/verticles/FakeS3BucketVerticle.java
```diff
@@ -37,13 +36,14 @@
...
-        final String deploymentID = deploymentID();
+    public void start() throws Exception {
+        super.start();
 
         LOGGER.debug(MessageCodes.MFS_110, getClass().getName(), deploymentID());
 
         if (myTmpDir == null) {
-            myTmpDir = Files.createTempDirectory(deploymentID).toFile();
+            // Creates a tmp dir from the deploymentID, adding additional random numbers after the underscore
+            myTmpDir = Files.createTempDirectory(deploymentID() + "_").toFile();
```
markmatney: Why this change?

markmatney: (the `_`)

ksclarke: The directory name is comprised of the prefix that I pass in (the deployment ID) and a random number. The underscore makes a clear delimiter that I can use to regexp match the first part of the dir name (the deployment ID) to find the directory.

src/test/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticleTest.java
ksclarke: The viewingHint and viewingDirection? We really only use the literals in the tests when they are repeated in the test (because our Checkstyle rules complain if we repeat a string more than once). We could put them in a TestConstants (at this point not sure I see a reason to put them into Constants since they're just used in a test) but since they're not repeated, I'd hold off until we have a need to do this.

markmatney: :+1:

src/test/java/edu/ucla/library/iiif/fester/verticles/FakeS3BucketVerticle.java
markmatney: That makes sense.

PR #54, opened 2020-01-17
 [IIIF-616] Delete uploaded CSV files after PostCsvHandler sends a response

markmatney:
- Set a custom BodyHandler on PostCsvHandler that ensures uploaded files are eventually deleted from disk
- Remove some redundant annotations on the handler test classes

(PostCsvHandlerTest has even more repeated code now, so it should be refactored at some point.)

### DIFF
--- a/src/test/java/edu/ucla/library/iiif/fester/handlers/PostCsvHandlerTest.java
+++ b/src/test/java/edu/ucla/library/iiif/fester/handlers/PostCsvHandlerTest.java
@@ -8,6 +8,8 @@ import java.io.File;
 import java.io.IOException;
 import java.io.StringReader;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.junit.After;
 import org.junit.Test;
@@ -32,6 +34,7 @@ import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.web.client.HttpRequest;
 import io.vertx.ext.web.client.HttpResponse;
 import io.vertx.ext.web.client.WebClient;
+import io.vertx.ext.web.handler.BodyHandler;
 import io.vertx.ext.web.multipart.MultipartForm;
 
 /**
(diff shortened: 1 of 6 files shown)


src/test/java/edu/ucla/library/iiif/fester/handlers/PostCsvHandlerTest.java
```diff
@@ -261,6 +264,74 @@ public final void testWorksCSV(final TestContext aContext) throws IOException, C
...
+
+                    // Check that the UUID-named CSV file was deleted
+                    final Matcher uploadedFilePathMatcher = Pattern
+                            .compile("(" + BodyHandler.DEFAULT_UPLOADS_DIRECTORY + "\\/" + "[0-9a-f\\-]+" + ")")
+                            .matcher(postStatusMessage);
+                    uploadedFilePathMatcher.find();
+                    final String uploadedFilePath = uploadedFilePathMatcher.group(1);
+
+                    // Wait for the file to get deleted; 500 ms should be long enough, unless the file is huge
+                    final Long timerDelay = 500L;
```
ksclarke: This is just a stylistic thing that I'll bring up for your thoughts. I tend to like defining variables at the top of a block (keeping them in a predictable place). I know others like defining them just before they're used (in case for some reason they don't end up needing to be used). A compromise might be to put the `final Long timerDelay;` under `final String contentType ...` and to initialize it where it is now with `timerDelay = 500;`. It's just a stylistic preference, but bringing it up to see if you have strong feelings about any of the three options.

src/test/java/edu/ucla/library/iiif/fester/handlers/PostCsvHandlerTest.java
```diff
@@ -261,6 +264,74 @@ public final void testWorksCSV(final TestContext aContext) throws IOException, C
...
+                    check(aContext, LinkUtils.addManifests(host, expected), actual);
+
+                    // Check that what we get back has the correct media type
+                    aContext.assertEquals(Constants.CSV_MEDIA_TYPE, contentType);
+
+                    // Check that the UUID-named CSV file was deleted
+                    final Matcher uploadedFilePathMatcher = Pattern
+                            .compile("(" + BodyHandler.DEFAULT_UPLOADS_DIRECTORY + "\\/" + "[0-9a-f\\-]+" + ")")
+                            .matcher(postStatusMessage);
+                    uploadedFilePathMatcher.find();
```
ksclarke: I know the expectation is that this will always be true, but wonder if we should put some logic in to handle if it ever returns false(?)

src/test/java/edu/ucla/library/iiif/fester/handlers/PostCsvHandlerTest.java
```diff
@@ -261,6 +264,74 @@ public final void testWorksCSV(final TestContext aContext) throws IOException, C
...
+                    final String uploadedFilePath = uploadedFilePathMatcher.group(1);
+
+                    // Wait for the file to get deleted; 500 ms should be long enough, unless the file is huge
+                    final Long timerDelay = 500L;
+                    myVertx.setTimer(timerDelay, timerId -> {
+                        try {
+                            final boolean isDeleted = !myVertx.fileSystem().existsBlocking(uploadedFilePath);
+                            aContext.assertTrue(isDeleted);
+                        } catch (AssertionError e) {
```
ksclarke: We've tended to use `details` as the default exception/error variable instead of the more traditional/jargon-y `e`. Not a huge deal, just mentioning it since I'm for more readable variables (another example would be `index` instead of `i`). Bucking tradition, I know. :-)

markmatney:
@ksclarke please review changes

PR #59, opened 2020-02-01
[IIIF-643] Implement GET for collection manifests

markmatney:
- Adds `GetCollectionHandler` to handle `GET` requests for collections
- Rename `AbstractManifestHandler` and its test class
- **Major refactor**: make the mapping between IDs (ARKs), IIIF resource URIs, and AWS S3 keys more explicit (49f422d):
  - Removed distinction between "JSON-less" and "JSON" S3 keys; all S3 keys now end with `.json`
  - Lots of local variable and member variable name changes, to make their contents more explicit (what *kind* of identifier is it: an ID (ARK), a URI (or a path), or an S3 key?)
  - All of the identifier mapping logic should now be in `IDUtils`
  - `AbstractFesterVerticle.getS3Manifest` signature changed to accept an `ObjecType`
  - `S3BucketVerticle.get` and `FakeS3BucketVerticle.get` Javadoc clarified that those methods accept an S3 key


src/main/java/edu/ucla/library/iiif/fester/utils/IDUtils.java
```diff
@@ -70,58 +68,106 @@ public static String getLastPart(final String aID) {
...
-     * @param aID The encoded work ID
-     * @return A decoded ID
+     * @param aID The ID (ARK) of the work
+     * @return An S3 key
      */
-    public static String decode(final URI aID) {
-        final String path = aID.getPath();
-        final int endIndex = path.lastIndexOf('/');
+    public static String getWorkS3Key(final String aID) {
+        return aID + Constants.JSON_EXT;
```
ksclarke: So this is a breaking change. It won't handle requests that come in with a JSON any more. Or would you prefer to move the logic that that checks for the .json and doesn't add it if it's there here?

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
```diff
@@ -581,7 +576,8 @@ private void extractWorkMetadata(final String[] aRow, final CsvHeaders aHeaders,
 
         // Create a brief work manifest for inclusion in the collection manifest
         if (id != null && label != null) {
-            final Collection.Manifest manifest = new Collection.Manifest(IDUtils.encode(myHost, id), label);
+            final Collection.Manifest manifest = new Collection.Manifest(IDUtils.getResourceURI(myHost, id).toString(),
```
ksclarke: At the point where the line is wrapping because of nested methods, I tend to think it's more readable to break it out. So, for instance, put `IDUtils.getResourceURI` into a `resourceURI` (or something) variable and pass that variable to new `Collection.manifest()`.

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
```diff
@@ -642,7 +638,8 @@ private Collection getCollection(final String[] aRow, final CsvHeaders aHeaders)
             final String label = StringUtils.trimToNull(aRow[aHeaders.getTitleIndex()]);
 
             if (label != null) {
-                return new Collection(IDUtils.encode(myHost, Constants.COLLECTIONS_PATH, id), label);
+                return new Collection(
+                        IDUtils.getResourceURI(myHost, IDUtils.getCollectionS3Key(id)).toString(), label);
```
ksclarke: At the point where the line is wrapping because of nested methods, I tend to think it's more readable to break it out. So, for instance, put `IDUtils.getCollectionS3Key` into a `s3Key` (or something) variable and pass that variable to `IDUtils.getResourceURI`.

src/test/java/edu/ucla/library/iiif/fester/handlers/GetCollectionHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+                    // Verify that the CORS header is permissive
+                    aContext.assertEquals(Constants.STAR, response.getHeader(Constants.CORS_HEADER));
+
+                    // Verify that our retrieved JSON is as we expect it
+                    aContext.assertEquals(new JsonObject(expectedCollection), new JsonObject(foundCollection));
+                    asyncTask.complete();
+                });
+            } else {
+                aContext.fail(LOGGER.getMessage(MessageCodes.MFS_003, HTTP.OK, statusCode));
+                asyncTask.complete();
```
ksclarke: A `aContext.fail` will complete the task so we've been wrapping the `asyncTask.complete()` calls in `if (!asyncTask.isCompleted()) { }` to avoid the warning message that comes out in the log when you complete a task that's already been completed.

src/test/java/edu/ucla/library/iiif/fester/handlers/GetCollectionHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+            if (response.statusCode() == HTTP.OK) {
+                response.bodyHandler(body -> {
+                    final String foundCollection = body.toString(StandardCharsets.UTF_8);
+
+                    // Verify that the CORS header is permissive
+                    aContext.assertEquals(Constants.STAR, response.getHeader(Constants.CORS_HEADER));
+
+                    // Verify that our retrieved JSON is as we expect it
+                    aContext.assertEquals(new JsonObject(expectedCollection), new JsonObject(foundCollection));
```
ksclarke: A `aContext.assertEquals` will complete the task when the assertion fails so in these situations we've been wrapping the `asyncTask.complete()` calls in `if (!asyncTask.isCompleted()) { }` to avoid the warning message that comes out in the log when you complete a task that's already been completed. This also applies to line 81.

src/test/java/edu/ucla/library/iiif/fester/handlers/GetManifestHandlerTest.java
```diff
@@ -102,9 +68,9 @@ public void testGetManifestHandlerJson(final TestContext aContext) throws IOExce
     public void testGetManifestHandler404(final TestContext aContext) {
         final Async asyncTask = aContext.async();
         final int port = aContext.get(Config.HTTP_PORT);
-        final String testIDPath = "/testIdentifier"; // path should be: /{id}/manifest
+        final String badPath = "/badIdentifier/manifest"; // path should be: /{id}/manifest
```
ksclarke: Perhaps `/missingIdentifier/manifest` or `/404Identifier/manifest`? The ID there isn't bad, it's just not resolvable.

src/main/java/edu/ucla/library/iiif/fester/utils/IDUtils.java
ksclarke: I guess this raises the question too of removing the tests related to confirming that .json ended IDs works. I'm more comfortable removing them now though and treating the support for ".json" requests as deprecated. We can officially remove it from here when we go to 1.x.x.

markmatney: Is this necessary though, since requests on `/{manifestId}/manifest.json` aren't covered by Fester's HTTP API spec, and the S3 API is internal? BTW, it looks like all the S3 keys currently on the prod bucket end with `.json`, so even that shouldn't break.

However, I now realize we will need to migrate the existing collections manifests from `/collections/` to `collections/`.

ksclarke: Isn't this being called from GetManifestHandler which handles Web requests (i.e. is outward facing)? There was logic there that's been removed that checked if the json extension was supplied. It didn't add the extension if it found it existed. It's not so much about what's on S3. A request with the json extension would now to go S3 with id.json.json. I realize we probably don't have any code any more that's supplying the json to the id passed to the handler (maybe?), but it is a public facing endpoint that we're changing (not in this exact location but by changing the GetManifestHandler to use this and removing the extension check from there -- line 44).

ksclarke: The change could be to GetManifestHandler so it does that check there, but I thought of putting it here since you're consolidating how ID operations are handled.

markmatney: Oh I see now, the `manifestId` param (not the `manifest` path component) could have a `.json` suffix. I agree on continuing to support but deprecating.

ksclarke: Yeah, maybe it just belongs in the GetManifestHandler still and keep the IDUtils clean and purposeful in what they're trying to do.

markmatney:
ready for review again @ksclarke

PR #65, opened 2020-02-12
[IIIF-644] Implement PUT for collection docs

markmatney:
In addition to the notes in the commits:
- I removed `AbstractFesterVerticle.getS3Manifest` (since it was only being called once) and replaced the call to it with the more generic `sendMessage`, but I'm not super opinionated on removing it if it's deemed useful.
- `FakeS3BucketVerticle`'s `JSON_FILES` is now mutable, so that it supports PUT.
- Note: all that's really going on at `@@ -92,72 +92,86 @@ public void start(final Promise<Void> aPromise) {` in the diff for `ManifestVerticle` is a change in indentation, since that code block is now inside a switch statement.

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandler.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandler.java
@@ -1,0 +1,60 @@
+package edu.ucla.library.iiif.fester.handlers;
+
+import edu.ucla.library.iiif.fester.Constants;
+import edu.ucla.library.iiif.fester.HTTP;
+import edu.ucla.library.iiif.fester.MessageCodes;
+import edu.ucla.library.iiif.fester.Op;
+import edu.ucla.library.iiif.fester.verticles.S3BucketVerticle;
+import info.freelibrary.util.Logger;
+import info.freelibrary.util.LoggerFactory;
+import io.vertx.core.Vertx;
+import io.vertx.core.eventbus.DeliveryOptions;
+import io.vertx.core.http.HttpServerResponse;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.RoutingContext;
+
+public class PutCollectionHandler extends AbstractFesterHandler {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(PutCollectionHandler.class, Constants.MESSAGES);
+
+    /**
+     * Creates a handler that creates IIIF collection manifests in Fester.
+     *
+     * @param aVertx A Vert.x instance
+     * @param aConfig A JSON configuration
+     */
+    public PutCollectionHandler(final Vertx aVertx, final JsonObject aConfig) {
+        super(aVertx, aConfig);
+    }
+
+    @Override
+    public void handle(final RoutingContext aContext) {
+        final HttpServerResponse response = aContext.response();
+        final String collectionName = aContext.request().getParam(Constants.COLLECTION_NAME);
+        final JsonObject message = new JsonObject();
+        final DeliveryOptions options = new DeliveryOptions();
+
+        message.put(Constants.COLLECTION_NAME, collectionName).put(Constants.DATA, aContext.getBodyAsJson());
+        options.addHeader(Constants.ACTION, Op.PUT_COLLECTION);
+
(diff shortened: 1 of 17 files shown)


src/main/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandler.java
```diff
@@ -0,0 +1,60 @@
...
+        sendMessage(S3BucketVerticle.class.getName(), message, options, send -> {
+            if (send.succeeded()) {
+                response.setStatusCode(HTTP.OK);
+                response.end();
+            } else {
+                final Throwable aThrowable = send.cause();
+                final String exceptionMessage;
+                final String errorMessage;
+
+                exceptionMessage = aThrowable.getMessage();
```
ksclarke: Why not put this with the variable definition on line 46?

src/main/java/edu/ucla/library/iiif/fester/verticles/S3BucketVerticle.java
```diff
@@ -68,31 +65,47 @@ public void start() throws Exception {
...
-                put(json, message);
+                    break;
+                case Op.PUT_COLLECTION:
+                    manifestID = msg.getString(Constants.COLLECTION_NAME);
+                    manifest = msg.getJsonObject(Constants.DATA);
+                    put(IDUtils.getCollectionS3Key(manifestID), manifest, message);
+                    break;
+                default:
+                    final String errorMessage = StringUtils.format(MessageCodes.MFS_139,
+                            this.getClass().toString(), message.toString(), action);
```
ksclarke: 'this' isn't needed here

src/main/resources/fester_messages.xml
```diff
@@ -131,4 +131,8 @@
   <entry key="MFS-135">Regular expression '{}' does not match '{}'</entry>
   <entry key="MFS-136">Cannot retrieve object '{}' from S3: unknown type '{}'</entry>
   <entry key="MFS-137">Test GetCollectionHandler with a test manifest: {}</entry>
+  <entry key="MFS-138">Supplied S3 key '{}' does not match what would have been derived from the IIIF resource URI: '{}'</entry>
+  <entry key="MFS-139">'{}' unable to handle message '{}': unknown action '{}'</entry>
+  <entry key="MFS-140">FAIL--manifest '{}' not found when it should have been put</entry>
```
ksclarke: "FAIL" seems redundant here since the message is displayed with a test failure. Same for the message below.

src/test/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandlerTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+        LOGGER.debug(MessageCodes.MFS_016, requestPath);
+
+        requestOpts.setPort(port).setHost(Constants.UNSPECIFIED_HOST).setURI(requestPath);
+        requestOpts.addHeader(Constants.CONTENT_TYPE, Constants.JSON_MEDIA_TYPE);
+
+        myVertx.createHttpClient().put(requestOpts, putResponse -> {
+            final int putStatusCode = putResponse.statusCode();
+
+            switch (putStatusCode) {
```
ksclarke: Not a big deal, just wondering why a `switch` here instead of an `if` / `else` with just the two choices?

src/main/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandler.java
markmatney: Good catch.

src/test/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandlerTest.java
markmatney: I think when I started editing this code, I had 3 case blocks, but now that there are only two it makes sense to use a simple if/else.

src/main/resources/fester_messages.xml
markmatney: The reason for it is that I was trying to be consistent with the style used in e.g. MFS-036, but I am happy to remove.

PR #70, opened 2020-02-27
[IIIF-729] Allow submission of CSVs with no "IIIF Access URL" column

markmatney:


markmatney:
Note that `ara249.csv` was moved under a directory to match the structure of our Sinai CSV repository.

markmatney:
The changes to the CSVs do pertain to the code changes because they make the tests fail.

markmatney:
I realize that I'm reading your comment slightly out of context now, though, since there were a lot more additions before I force-pushed (many of which did not pertain to the code changes).

PR #77, opened 2020-03-25
Update status API endpoint

ksclarke:
* Update status endpoint path
* Update documentation about the status endpoint
* Remove obsolete MainVerticleTest which, despite its name, was a test of the status endpoint (which we already do in our new FIT tests).


markmatney:
(and the main README.md, too)

PR #79, opened 2020-03-29
[IIIF-744] Create feature flag for batch ingests

ksclarke:
* Add a feature flag for CSV batch ingests
* Add placeholder page for turned-off batch ingest feature
* Add two docker spin up tests for the feature flags
  * One with the flag turned on
  * Another with the flag turned off
* Update S3BucketVerticle to be able to use localstack s3 for functional testing
* Update logging configs
* Move fester configs into /etc/fester dir instead of just /etc
* Updated the deprecated methods in MainVerticle (while refactoring for feature flags)

In addition to FT (functional tests), UT (unit tests), and IT (integration tests), we now have FfT (feature flag tests). FfTs are run in the integration Maven phase, like the other Docker-based tests.


src/main/java/edu/ucla/library/iiif/fester/handlers/EndpointConfigHandler.java
```diff
@@ -0,0 +1,198 @@
...
+            // We need to associate endpoint handlers with routes from our specification
+            factory.addHandlerByOperationId(Op.GET_STATUS, new GetStatusHandler());
+            factory.addHandlerByOperationId(Op.GET_MANIFEST, new GetManifestHandler(myVertx, myConfig));
+            factory.addHandlerByOperationId(Op.PUT_MANIFEST, new PutManifestHandler(myVertx, myConfig));
+            factory.addHandlerByOperationId(Op.DELETE_MANIFEST, new DeleteManifestHandler(myVertx, myConfig));
+            factory.addHandlerByOperationId(Op.GET_COLLECTION, new GetCollectionHandler(myVertx, myConfig));
+            factory.addHandlerByOperationId(Op.PUT_COLLECTION, new PutCollectionHandler(myVertx, myConfig));
+
+            // After the batch ingest feature is configured (or not), we complete the router configuration
+            featureConfigPromise.future().setHandler(handler -> {
```
markmatney: The only reason I'm raising this is because I think it hinders code readability -- it took me a little while to wrap my head around what `featureConfigPromise` was for. I think it would be clearer to instead fail `myPromise` straight from `configureBatchIngest`.

src/main/java/edu/ucla/library/iiif/fester/handlers/EndpointConfigHandler.java
```diff
@@ -0,0 +1,198 @@
...
+        } else {
+            myPromise.fail(aConfiguration.cause());
+        }
+    }
+
+    /**
+     * Configure the router with additional paths that are not handled by the OpenAPI router factory
+     *
+     * @param aFactory An OpenAPI router factory
+     * @param aPromise A router configuration promise
```
markmatney: `aPromise` isn't passed here

src/main/java/edu/ucla/library/iiif/fester/handlers/EndpointConfigHandler.java
```diff
@@ -0,0 +1,198 @@
...
+        } catch (final IOException details) {
+            aPromise.fail(details);
+        }
+    }
+
+    /**
+     * Configure a placeholder for the ingest endpoint
+     *
+     * @param aFactory A router factory
+     * @param aConfig An application configuration
```
markmatney: `aConfig` isn't passed here

src/test/java/edu/ucla/library/iiif/fester/fit/BaseFesterFfT.java
```diff
@@ -0,0 +1,26 @@
...
+public class BaseFesterFfT extends BaseFesterFT {
+
+    /* The human friendly name of the batch ingest feature */
+    protected static final String BATCH_INGEST_FEATURE = Features.getDisplayName(Features.BATCH_INGEST);
+
+    /* The path at which our CSV upload form can be found */
+    protected static final String UPLOAD_FORM_PATH = "/fester/upload/csv";
+
+    /* The POST collections endpoint */
+    protected static final String API_PATH = "/collections";
```
markmatney: redefinition of `Constants.POST_CSV_ROUTE`

ksclarke: Good point.

src/main/java/edu/ucla/library/iiif/fester/handlers/EndpointConfigHandler.java
markmatney: Turns out I was mistaken, it looks like my proposed change wouldn't be an improvement. Thanks @ksclarke for walking me through

PR #82, opened 2020-04-07
[IIIF-464] Add human friendly responses to endpoints

ksclarke:
* Add human friendly text to the responses of our endpoints (that don't already output something like HTML that's intended for the user)
* Minor refactor on a variable name for consistency on naming standards
* Bump up the timeout on one of the tests that fails periodically because the S3BucketVerticle/FakeS3BucketVerticle swap doesn't happen in a timely manner
* Update OpenAPI spec with response content types and examples

### DIFF
--- a/src/main/resources/fester_messages.xml
+++ b/src/main/resources/fester_messages.xml
@@ -101,6 +101,14 @@
   <entry key="MFS-085">The '{}' feature is unavailable on this server</entry>
   <entry key="MFS-086">Turning the '{}' feature off</entry>
   <entry key="MFS-087">Unable to find the h1 element in the feature flag test</entry>
+  <entry key="MFS-088">Successfully deleted: {}</entry>
+  <entry key="MFS-089">Access forbidden: {}</entry>
+  <entry key="MFS-090">Unexpected response code [{}]: {}</entry>
+  <entry key="MFS-091">Resource not found: {}</entry>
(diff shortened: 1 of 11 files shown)


src/main/resources/fester_messages.xml
```diff
@@ -101,6 +101,14 @@
...
   <entry key="MFS-086">Turning the '{}' feature off</entry>
   <entry key="MFS-087">Unable to find the h1 element in the feature flag test</entry>
+  <entry key="MFS-088">Successfully deleted: {}</entry>
+  <entry key="MFS-089">Access forbidden: {}</entry>
+  <entry key="MFS-090">Unexpected response code [{}]: {}</entry>
+  <entry key="MFS-091">Resource not found: {}</entry>
+  <entry key="MFS-092">Manifest successfully PUT: {}</entry>
+  <entry key="MFS-093"></entry>
+  <entry key="MFS-094"></entry>
+  <entry key="MFS-095"></entry>
```
markmatney, line 111: remove since unused?

ksclarke: I'd just say leave for now. They're placeholders that save the time of having to run `rb` to add the new MessageCodes when they are used.

markmatney: ok

PR #83, opened 2020-04-07
[IIIF-788] Clean up obsolete Docker networks

ksclarke:
TestContainers seems to leave networks around despite Ryuk supposedly being responsible for cleaning them up. We add an additional cleanup as a part of our build so we don't get to the point where all Docker's networks are used up and we can't run the build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -330,6 +330,28 @@
           </execution>
         </executions>
       </plugin>
+      <!-- Makes sure TestContainers doesn't leave unused networks around -->
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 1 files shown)


ksclarke:
> Just wanted to flag that this could be potentially undesirable or dangerous, since it removes all unused networks on the Docker host, not just those used by the TestContainers (I think?). I think that's fine with me in this context, since I'm not doing any special Docker networking stuff on this machine. There are probably edge cases but I'm not sure those are worth worrying about.

Yes, it's only _unused_ networks though. So any networks connected to Docker containers should be left alone. I thought that was okay.

markmatney:
Ok, cool. I think it's fine :)

PR #84, opened 2020-04-07
[IIIF-673] Remove deprecated IDUtils method

ksclarke:


ksclarke:
@markmatney I had a merge conflict on this one so it reset your review. That will need to be done again, sorry.

PR #87, opened 2020-04-08
[IIIF-697] Improve error handling in IDUtils

ksclarke:
* Create new runtime exception for path (s3 or URI) problems
* Catch more possible (bad) variations in ID values
* Add new tests for those error checks
* Refactor tests for brevity

This PR is going to have conflicts with the one that removed the deprecated method from IDUtils, I fear.


ksclarke:
@markmatney I had to resolve a merge conflict so this needs to be re-reviewed.

PR #90, opened 2020-04-10
[IIIF-802] Suppress warnings from AWS SDK during Maven build

markmatney:
To test, create or modify `~/.aws/config` with something like:
```
[profile iiif_802]
region = us-west-1
output = json
```
and then run `mvn clean test`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -170,12 +170,6 @@
       <artifactId>vertx-web-api-contract</artifactId>
       <version>${vertx.version}</version>
     </dependency>
-    <!-- Allows us to suppress warnings from the AWS SDK (see https://github.com/aws/aws-sdk-java/issues/1707) -->
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>jcl-over-slf4j</artifactId>
-      <version>${jcl.over.slf4j.version}</version>
-    </dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -169,6 +170,12 @@
       <artifactId>vertx-web-api-contract</artifactId>
       <version>${vertx.version}</version>
     </dependency>
+    <!-- Allows us to suppress warnings from the AWS SDK (see https://github.com/aws/aws-sdk-java/issues/1707) -->
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>jcl-over-slf4j</artifactId>
+      <version>${jcl.over.slf4j.version}</version>
```
ksclarke, line 177: I think we should add a `<scope>test</scope>` so that the dependency won't be packaged into the final application jar. It's useful in our test environment, but I don't think we need it in the runtime? Also, with the next version of vertx-super-s3 the AWS dependency will also be removed (i.e., marked as a test dependency only).

ksclarke: When we put the scope on it we could move the dependency below the ` <!-- Below dependencies only used for testing -->` line in the pom.xml too.

markmatney:
Ok, ready for a second look

PR #91, opened 2020-04-13
[IIIF-745] Add base URL configuration option

markmatney:
Changes:
* update `GetManifestHandler` to use `S3BucketVerticle` instead of the S3 client
* update `S3BucketVerticle`
  * store manifests with all IIIF Presentation API resource URLs prefixed with a placeholder base URL                                                      
  * by default, return manifests with all instances of the placeholder base URL replaced with the actual base URL
* move `PostCsvHandler` tests to the FIT framework
* update URLs in the test fixtures
* update tests accordingly
* remove erroneous HTTP 400 entry in OpenAPI spec

Note: declarations of `myUrl` and `myUrlPlaceholderPattern` are duplicated a lot more than I'd like, but I expect this to change as we adopt the new FIT framework.


pom.xml
```diff
@@ -119,6 +119,9 @@
     <!-- The IIIF base URL (including service prefix) -->
     <iiif.base.url>https://iiif.library.ucla.edu/iiif/2</iiif.base.url>
 
+    <!-- The URL Fester is available at -->
+    <fester.url>https://iiif.library.ucla.edu</fester.url>
```
ksclarke, line 123: Shouldn't the default value be localhost for dev/testing? We'd override on deployed versions?

pom.xml
```diff
@@ -555,6 +558,7 @@
           </excludes>
           <systemPropertyVariables>
             <vertx.logger-delegate-factory-class-name>io.vertx.core.logging.SLF4JLogDelegateFactory</vertx.logger-delegate-factory-class-name>
+            <fester.url>http://localhost</fester.url>
```
ksclarke, line 561: I think the default should be localhost, making this unnecessary.

pom.xml
```diff
@@ -637,6 +641,7 @@
               <skipTests>${skipFTs}</skipTests>
               <systemPropertyVariables>
                 <fester.logs.output>${seeLogsFT}</fester.logs.output>
+                <fester.url>http://localhost</fester.url>
```
ksclarke, line 644: I think the default should be localhost, making this unnecessary.

src/main/java/edu/ucla/library/iiif/fester/Constants.java
```diff
@@ -112,6 +112,17 @@
...
+    /**
+     * Used as a message header when the sender wants the manifest returned from S3 without resource URLs re-written.
+     */
+    public static final String NO_REWRITE_URLS = "no-rewrite-urls";
+
+    /**
+     * A unique random placeholder URL that prefixes all IIIF Presentation API resource URLs in all manifests at rest in
+     * S3. It gets replaced with Constants.URL on each GET request.
+     */
+    public static final String URL_PLACEHOLDER = "http://b1dbe4a0-443c-479f-bf0a-25c352df0d8f.example.org";
```
ksclarke, line 124: Maybe use a UUID + a UCLA library domain (e.g. http://b1dbe4a0-443c-479f-bf0a-25c352df0d8f.iiif.library.ucla.edu)? I know the example.org domain is for examples, but I'm just a little hesitant to use it for a placeholder since we don't own it.

pom.xml
markmatney: My inclination was to use prod like `iiif.base.url`, since `fester.url` isn't dereferenced anywhere anyway, but this makes sense. Will fix.

src/main/java/edu/ucla/library/iiif/fester/handlers/GetCollectionHandler.java
```diff
@@ -52,8 +52,8 @@ public void handle(final RoutingContext aContext) {
             } else {
                 final Throwable aThrowable = send.cause();
                 final String exceptionMessage = aThrowable.getMessage();
-                final String errorMessage = LOGGER.getMessage(MessageCodes.MFS_009, collectionName,
-                        HTTP.INTERNAL_SERVER_ERROR, exceptionMessage);
+                final String errorMessage = LOGGER.getMessage(MessageCodes.MFS_009, collectionName, HTTP.NOT_FOUND,
```
ksclarke, line 55: So, I see we have an inconsistency here, but I'm not sure changing the exception message to a 404 accurately captures what's happened. Looking at our S3BucketVerticle, the `get` basically just checks for `get.statusCode() == HTTP.OK` and, if it doesn't find that, fails the response. That failure though doesn't really make a distinction between a failed because it's a 404 and failed because there was some sort of other failure (500, etc.) I'm wondering if the fail int code from the S3BucketVerticle shouldn't be the HTTP response code instead of the MessageCodes' int code (which is kind of useless since that particular message code's value is just `{}`). It seems like that failure int code should be actionable if we can set it.

markmatney:
still have a couple more commits incoming

markmatney:
Ok @ksclarke ready for a second look.

On the Docker tests: I know it's not ideal that FESTER_URL doesn't contain the actual port that the container is listening on, but I would need more time to figure out how to make that change. I was hoping that for now we could note this a bug and address it in the future. It should be fine for now, since we don't currently have any tests that dereference that URL.

I'm also now realizing that I may have read more into [your comment above](https://github.com/UCLALibrary/fester/pull/91#discussion_r407619129) when I added HTTP 500 to the OpenAPI spec in those two places; this is easy to remove if you wish.

ksclarke:
> On the Docker tests: I know it's not ideal that FESTER_URL doesn't contain the actual port that the container is listening on, but I would need more time to figure out how to make that change. I was hoping that for now we could note this a bug and address it in the future. It should be fine for now, since we don't currently have any tests that dereference that URL.

Yes, this is fine with me. Would you create the bug ticket and reference it here so we have a documentation trail?

ksclarke:
This branch is out of date with master and needs to be updated before it can be merged.

markmatney:
Okay, ready for re-review. The bug ticket is IIIF-804.

pom.xml
```diff
@@ -637,7 +639,6 @@
               <skipTests>${skipFTs}</skipTests>
               <systemPropertyVariables>
                 <fester.logs.output>${seeLogsFT}</fester.logs.output>
-                <fester.http.port>8888</fester.http.port>
```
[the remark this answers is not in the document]
markmatney: It's not required, I meant to remove it but missed it.

PR #94, opened 2020-04-16
[IIIF-803] Add procedure for installing Festerize with Pip

markmatney:
I will ask DLP to review.

### DIFF
--- a/src/main/scripts/festerize/README.md
+++ b/src/main/scripts/festerize/README.md
@@ -4,7 +4,7 @@ Uploads CSV files to the Fester IIIF manifest service.
 
 ## Installation
 
-First, ensure that you have Python 3 and Pip installed on your system.
+First, ensure that you have Bash, cURL, Python 3 and Pip installed on your system.
 
 When that's done, run the following command in your shell to install the latest release of Festerize:
 
--- a/src/main/scripts/festerize/README.md
+++ b/src/main/scripts/festerize/README.md
@@ -8,7 +8,7 @@ First, ensure that you have Python 3 and Pip installed on your system.
 
 When that's done, run the following command in your shell to install the latest release of Festerize:
 
-    bash <(curl -sSL https://raw.githubusercontent.com/UCLALibrary/fester/master/install.sh)
+    bash <(curl -sSL https://raw.githubusercontent.com/UCLALibrary/fester/master/src/main/scripts/festerize/install.sh)
 
 ## Usage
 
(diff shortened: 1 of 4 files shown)


src/main/scripts/festerize/README.md
```diff
@@ -4,30 +4,11 @@ Uploads CSV files to the Fester IIIF manifest service.
...
-If you want to install it in a place that's available to all users on a system (and you have sudo privileges) you can use:
-
-    sudo ./setup.py
-
-### Installing on a Mac
-
-To install on a Mac that's installed Python 3 with Homebrew, type:
-
-    python3 setup.py install
+    bash <(curl -sSL https://raw.githubusercontent.com/UCLALibrary/fester/master/install.sh)
```
ksclarke: Is this path right? It should be the `src/main/scripts/festerize` path right?

src/main/scripts/festerize/install.sh
```diff
@@ -0,0 +1,24 @@
...
+    if ! [ -x "$(command -v ${PIP})" ]; then
+        echo "${PIP} not installed" >&2
+        PIP=
+    else
+        break
+    fi
+done
+
+if [ -z ${PIP} ]; then
+    echo "A version of Pip compatible with Python 3 is required to install Festerize."
```
ksclarke, line 14: Does the above actually check that `pip` is v3? Are there systems for which `pip` might be pointed to a 2.x version?

src/main/scripts/festerize/README.md
```diff
@@ -4,30 +4,11 @@ Uploads CSV files to the Fester IIIF manifest service.
 
 ## Installation
 
-You need to have Python 3 installed to use Festerize. This README will describe the simplest way to get started; if you're more experienced and would like to use a virtualized Python environment, feel free to do that.
+First, ensure that you have Python 3 and Pip installed on your system.
```
ksclarke, line 7: And curl and bash, right?

src/main/scripts/festerize/README.md
markmatney: yeah I botched it, just pushed another commit that fixes this

src/main/scripts/festerize/README.md
markmatney: ah, true

src/main/scripts/festerize/install.sh
ksclarke: It could parse `pip -V` output perhaps?

markmatney: Yeah, I think that will work. I verified that it works on a Mac too. Just pushed a new commit.

PR #104, opened 2020-05-12
[IIIF-811] Allow non-EOL whitespace in CSV column headers

markmatney:


markmatney:
I've added a new test fixture with column headers that contain a mix of tabs and spaces.

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
```diff
@@ -67,7 +68,10 @@ public CsvParser parse(final Path aPath) throws IOException, CsvException, CsvPa
                 checkForEOLs(row);
 
                 if (myCsvHeaders == null) {
-                    myCsvHeaders = new CsvHeaders(row); // CsvParsingException if a 'required' header is missing
+                    final String[] headerRow = Arrays.asList(row).stream()
```
ksclarke, line 71: Isn't trimming to null is going to open the possibility of a NullPointerException in CsvHeaders (when it's passed to the `switch` in CsvHeaders' constructor)?

I also wonder whether we really need to create the extra List and second array when we could just do something that would use the existing array like:
```
for (int index = 0; index < row.length; index++) {
    row[index] = row[index].trim();
}
```

markmatney: Ah, I didn't realize a null value would cause a problem there. I see that it does. Will fix

markmatney:
Ready for re-review.

PR #106, opened 2020-05-19
Add multi-work page processing

ksclarke:
* Add `logLevel` to the Travis build so we can configure the log level of the container we build.
* Change `testLogLevel` to `logLevel` since it sets the log level in the deployed logging config too currently.
* Change the logging config in the "live" and "debug" profiles so they can use a local (in $basedir) logging configuration (as an override to what's in the repo). We did this for regular tests already... just extending it to the live test builds now too.
* Add more informational logging
* Make some Optional vars more obviously opts (so it's clearer one needs to do a `get()` on them)
* Handle an uncaught exception that happens when setting statusCode to -1 (which means the browser cancelled the request before the response was returned).
* Update to be able to rewrite the work manifest with the new pages being processed instead of adding to the existing manifest.
* Update to be able to process the pages of more than one work in a single ingest process where just a pages CSV is being uploaded.


src/main/java/edu/ucla/library/iiif/fester/handlers/GetManifestHandler.java
```diff
@@ -54,14 +54,20 @@ public void handle(final RoutingContext aContext) {
...
-                        statusMessage);
+                final String errorMessage;
 
+                int responseCode = failure.failureCode();
+
+                errorMessage = LOGGER.getMessage(MessageCodes.MFS_009, manifestID, responseCode, statusMessage);
                 LOGGER.error(errorMessage);
 
-                response.setStatusCode(statusCode);
+                // If the browser stops the request we'll get back a -1 code which causes setStatusCode to error
```
markmatney, line 65: interesting, good catch

PR #107, opened 2020-05-20
[IIIF-849] Handle uploaded CSV with blank lines

ksclarke:
* Add support for parsing CSV files with blank lines
* Fix misconfiguration in logging configuration file
* Remove some boilerplate code from PostCsvFIT


src/test/java/edu/ucla/library/iiif/fester/fit/PostCsvFIT.java
```diff
@@ -143,52 +165,45 @@ public final void testFullCSV(final TestContext aContext) throws CsvException, I
...
          */
         @Test
-        @SuppressWarnings("checkstyle:indentation")
         public final void testFullCsvWithIiifHost(final TestContext aContext) {
             final Async asyncTask = aContext.async();
 
-            final String filename = ALL_IN_ONE_CSV.getName();
-            final String pathname = ALL_IN_ONE_CSV.getAbsolutePath();
-            final MultipartForm form = MultipartForm.create().textFileUpload(Constants.CSV_FILE, filename, pathname,
-                    Constants.CSV_MEDIA_TYPE).attribute(Constants.IIIF_HOST, ImageInfoLookup.FAKE_IIIF_SERVER);
```
markmatney, line 153: Not all of the tests had been using a form with the `IIIF_HOST` attribute set, but now they are. Is that intentional?

ksclarke: Not all of them needed that because not all tested images. It doesn't hurt if they have it, though, even if they don't absolutely need it. It would always be supplied by the form if someone was manually running the tests.

markmatney: Want to remove `testFullCsvWithIiifHost` then? It's now identical to `testFullCSV`.

markmatney: Also it looks like the `throws` declarations on `testFullCSV` are no longer needed.

ksclarke: Okay, done. Good catch.

PR #111, opened 2020-06-03
[IIIF-845] Sort manifests by label in collections manifest

markmatney:
- Implement comparator for manifests based on their labels
  - This uses an additional third-party dependency that implements alpha-numeric sorting like I think we want. I have reviewed that dependency's source, and it only depends on the Java standard library and looks safe to use.
- Rewrite `ManifestVerticle.updateCollection` to use the comparator


markmatney:
btw, here's the repository for that third-party code https://github.com/sawano/alphanumeric-comparator

src/main/java/edu/ucla/library/iiif/fester/utils/ManifestLabelComparator.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Creates an Manifest Label comparator.
+     */
+    public ManifestLabelComparator() {
+    }
+
+    @Override
+    public int compare(final Collection.Manifest aFirstManifest, final Collection.Manifest aSecondManifest) {
+        // Each label is a list of values
+        final Iterator<Value> firstLabelValues = aFirstManifest.getLabel().getValues().iterator();
```
ksclarke, line 27: Hmm, so this assumes the order is set with the most important label first? The ticket just seems to assume maybe that there is only one label or that maybe the most important is first? I wonder what happens when we have multiple labels from different languages (each saying the same thing in a different language, like we might have with the Sinai manifests). I wonder if we should clarify with Dawn whether first label is the most appropriate label or whether we'd want to pass a configurable language code into the comparator so the sorting could be done in a language-preferred way?

src/test/java/edu/ucla/library/iiif/fester/fit/PostCsvFIT.java
```diff
@@ -356,6 +363,72 @@ public final void testDeleteUploadedFilesOnEnd(final TestContext aContext) throw
...
+
+                                    for (int i = 0; i < works.size() - 1; i++) {
+                                        aContext.assertTrue(comparator.compare(works.get(i), works.get(i + 1)) < 0);
+                                    }
+
+                                    if (!asyncTask.isCompleted()) {
+                                        asyncTask.complete();
+                                    }
+                                } else {
+                                    aContext.fail(
```
ksclarke, line 410: Wondering about shortening collectionS3Key to just `s3Key` since we're getting pretty far over to the right side (and wrapping). The method is dealing with collections so think `s3Key` would be sufficient.

src/main/java/edu/ucla/library/iiif/fester/utils/ManifestLabelComparator.java
markmatney: This is a good point. I made some assumptions here that I shouldn't have. I can follow up with Dawn on this.

markmatney:
After discussing with Dawn, I realized that a much simpler implementation of ManifestLabelComparator is sufficient.

Ready for re-review.

src/main/java/edu/ucla/library/iiif/fester/utils/ManifestLabelComparator.java
```diff
@@ -0,0 +1,51 @@
...
+import se.sawano.java.text.AlphanumericComparator;
+
+/**
+ * Comparator for sorting manifests alpha-numerically based on their label.
+ */
+public class ManifestLabelComparator implements Comparator<Collection.Manifest> {
+
+    private final AlphanumericComparator myComparator = new AlphanumericComparator();
+
+    /**
```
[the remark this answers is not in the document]
markmatney: I keep forgetting this. Noted.

PR #115, opened 2020-06-09
[IIIF-828] Improve Festerize description

### LINKED ISSUE #102: Clarify Readme.md
For me the use of the word "default" in the readme (and in the help) implies that if you like the default, then you don't need to specify. Instead, in this instance I think the better word might be "expected" or "most usual".

markmatney:
fixes #102

### DIFF
--- a/src/main/scripts/festerize/README.md
+++ b/src/main/scripts/festerize/README.md
@@ -1,10 +1,16 @@
 # Festerize
 
-Uploads CSV files to the Fester IIIF manifest service.
+Uploads CSV files to the Fester IIIF manifest service for processing.
 
-After Fester processes the CSVs in order to create or update any IIIF collections or manifests, it updates and returns those CSV files to the user.
+Any collection rows found in the CSV are used to create a IIIF collection.
 
-The returned CSVs are updated to contain URLs of the IIIF collections and (work) manifests that correspond to any collection or work rows found in the CSV.
+Any work rows are used to expand or revise a previously created IIIF collection (corresponding to the collection that the work is a part of), as well as create a IIIF manifest corresponding to the work.
+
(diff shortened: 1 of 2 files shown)


src/main/scripts/festerize/README.md
```diff
@@ -2,6 +2,10 @@
 
 Uploads CSV files to the Fester IIIF manifest service.
 
+After Fester processes the CSVs in order to create or update any IIIF collections or manifests, it updates and returns those CSV files to the user.
```
emcaulay, line 5: If the CSV contains a work that has never been festerized before, then a column is added to the CSV that provides the IIIF Manifest URL. If the CSV contains pages, the values in the CSV are used to expand or revise a previously festerized work record.

emcaulay: That's my understanding of what happens. I know you have that in the next paragraph to some extent so you might want to revise to merge my comment with your description.

src/main/scripts/festerize/README.md
```diff
@@ -24,6 +28,12 @@ Usage: festerize [OPTIONS] SRC...
 
   Uploads CSV files to the Fester IIIF manifest service.
```
emcaulay, line 29: I don't understand this paragraph and how it relates to the beginning of the file. It is possible I'm misreading the Github display. But I think this repeats the same heading as earlier in the readme.

markmatney: This is just a copy of what's at the beginning of the file (as well as the description in the Python script itself).

At some point we decided to put the entire "help" output inside the README, but whether or not that is useful is open for discussion.

ksclarke: Yes, that's a question we should ask again... @emcaulay, we could just have the read me say, "See help output for use"... Is there value putting documentation on how to use the script in the README or would you prefer just a pointer to the "help" output?

PR #117, opened 2020-06-12
[IIIF-863] Improve "manifest not found" error message

markmatney:
- if `getLockedManifest` fails to retrieve a manifest, send the ARK and the manifest type back to the client
- assign an id to the error message HTML tag
- on the client running `festerize`, only log the error message, not the entire HTML response

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
@@ -331,7 +331,7 @@ public class ManifestVerticle extends AbstractFesterVerticle {
                             final String errorMessage = LOGGER.getMessage(MessageCodes.MFS_146,
                                     aCollDoc ? "collection" : "work", aID);
                             lockRequest.result().release();
-                            aPromise.fail(new Throwable(errorMessage));
+                            aPromise.fail(new Exception(errorMessage, handler.cause()));
                         }
                     });
                 } catch (final NullPointerException | IndexOutOfBoundsException details) {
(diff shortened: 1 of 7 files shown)


src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
```diff
@@ -328,8 +328,10 @@ private void getLockedManifest(final String aID, final boolean aCollDoc, final P
 
                             aPromise.complete(new LockedManifest(manifest, aCollDoc, lock));
                         } else {
+                            final String errorMessage = LOGGER.getMessage(MessageCodes.MFS_146,
+                                    aCollDoc ? "collection" : "work", aID);
                             lockRequest.result().release();
-                            aPromise.fail(handler.cause());
+                            aPromise.fail(new Throwable(errorMessage));
```
ksclarke, line 334: It would be better to create an exception here that extends an Exception rather than use a generic Throwable. You can also give that exception the handler.cause() exception as its parent exception so that we don't lose that stacktrace.

src/main/scripts/festerize/festerize.py
```diff
@@ -84,10 +85,10 @@ def cli(src, server, endpoint, out, iiifhost, loglevel):
                 out_file = click.open_file(os.path.join(out, csv_filename), 'wb')
                 out_file.write(r.content)
             else:
-                error_msg = 'Failed to upload {}: {}'.format(csv_filename, r.status_code)
+                error_cause = BeautifulSoup(r.text, features='html.parser').find(id='error-message').string
```
ksclarke, line 88: Nice improvement!

markmatney:
I see, I'll make that change

PR #119, opened 2020-06-19
added a definition and a correction on line 11

emcaulay:
Also, I'm testing whether this is a viable workflow for me to provide comments / edits.

### DIFF
--- a/src/main/scripts/festerize/README.md
+++ b/src/main/scripts/festerize/README.md
@@ -4,11 +4,11 @@ Uploads CSV files to the Fester IIIF manifest service for processing.
 
 Any collection rows found in the CSV are used to create a IIIF collection.
 
-Any work rows are used to expand or revise a previously created IIIF collection (corresponding to the collection that the work is a part of), as well as create a IIIF manifest corresponding to the work.
+Any work rows are used to expand or revise a previously created IIIF collection (corresponding to the collection that the work is a part of), as well as create a IIIF manifest corresponding to the work. A "work" is conceived of as a discrete object that should be access as an individual item. An example of a "work" is a book or a photograph. 
 
-Any page rows are likewise used to expand or revise a previously created IIIF manifest (corresponding to the work that the page is a part of).
+Any page rows found in a CSV are likewise used to expand or revise a previously created IIIF manifest (corresponding to the work that the page is a part of).
 
-After Fester creates or updates any IIIF collections or manifests, it updates and returns the CSV files to the user.
+After Fester creates or updates any IIIF manifests for collections or works, it updates and returns the CSV files to the user.
 
 The returned CSVs are updated to contain URLs (in a `IIIF Manifest URL` column) of the IIIF collections and manifests that correspond to any collection or work rows found in the CSV.
 


src/main/scripts/festerize/README.md
```diff
@@ -4,11 +4,11 @@ Uploads CSV files to the Fester IIIF manifest service for processing.
 
 Any collection rows found in the CSV are used to create a IIIF collection.
 
-Any work rows are used to expand or revise a previously created IIIF collection (corresponding to the collection that the work is a part of), as well as create a IIIF manifest corresponding to the work.
+Any work rows are used to expand or revise a previously created IIIF collection (corresponding to the collection that the work is a part of), as well as create a IIIF manifest corresponding to the work. A "work" is conceived of as a discrete object that should be access as an individual item. An example of a "work" is a book or a photograph.
```
ksclarke, line 7: _accessed_

`can be accessed` instead of should?

emcaulay: Good edit!

src/main/scripts/festerize/README.md
```diff
@@ -4,11 +4,11 @@ Uploads CSV files to the Fester IIIF manifest service for processing.
 
 Any collection rows found in the CSV are used to create a IIIF collection.
 
-Any work rows are used to expand or revise a previously created IIIF collection (corresponding to the collection that the work is a part of), as well as create a IIIF manifest corresponding to the work.
+Any work rows are used to expand or revise a previously created IIIF collection (corresponding to the collection that the work is a part of), as well as create a IIIF manifest corresponding to the work. A "work" is conceived of as a discrete object that should be access as an individual item. An example of a "work" is a book or a photograph. 
 
-Any page rows are likewise used to expand or revise a previously created IIIF manifest (corresponding to the work that the page is a part of).
+Any page rows found in a CSV are likewise used to expand or revise a previously created IIIF manifest (corresponding to the work that the page is a part of).
```
markmatney, line 9: RE: your comment in Jira @emcaulay -- yes, this wording is incorrect. Currently, if you give Fester a CSV with both work rows and page rows, it will ignore the page rows. If you want it to do something with the page rows, it must not contain any work rows (or collection rows, for that matter). I will fix this to correctly describe the current behavior.

src/main/scripts/festerize/README.md
```diff
@@ -4,11 +4,11 @@ Uploads CSV files to the Fester IIIF manifest service for processing.
...
 
-After Fester creates or updates any IIIF collections or manifests, it updates and returns the CSV files to the user.
+After Fester creates or updates any IIIF manifests for collections or works, it updates and returns the CSV files to the user.
```
markmatney, line 11: We discussed on Slack and decided to keep the original wording, to conform with the IIIF terminology.

src/main/scripts/festerize/README.md
ksclarke: If you give Fester a CSV with Collection, Work, and Page rows, it should process all.

emcaulay:
We discussed all these revisions. My revisions aren't quite correct, and it helped to discuss the topics, but the pull request is not needed at this time. Mark will do more revisions to incorporate our discussion.

src/main/scripts/festerize/README.md
markmatney: Okay, I must be reading the code wrong then. I think you're right @ksclarke and I think my first comment is incorrect.

PR #121, opened 2020-07-01
[IIIF-828] Improve Festerize and Fester description

markmatney:
This incorporates changes we've discussed in Jira, and in #115 and #119.


markmatney:
We've decided in standup today that this doesn't need an additional review.

PR #122, opened 2020-07-09
Iiif 846

DRickard:
Added a method to trim leading/trailing white space in non-header CSV entries
Added a test for new code
Added a CSV for the new test


DRickard:
Cancelling this pull request; will fix code style and open a new one.

PR #123, opened 2020-07-13
Iiif 846

DRickard:
Updates:

- Added method to clear whitespace in non-header rows
- Added test for new method
- Added CSV for new test
- Added myself to "developers" block in pom


src/test/java/edu/ucla/library/iiif/fester/CsvParserTest.java
```diff
@@ -45,6 +45,18 @@ public final void testParse() throws CsvParsingException, CsvException, IOExcept
...
+    /**
+     * Tests handling spaces in non-header rows in a CSV file.
+     *
+     * @throws CsvParsingException If there is an error while parsing the CSV data
+     * @throws CsvException If there is a generic CSV error
+     * @throws IOException If there is trouble reading the CSV data
+     */
+    @Test
+    public final void testParseSpaces() throws CsvParsingException, CsvException, IOException {
+        myCsvParser.parse(getTestPath("spaces.csv"));
```
ksclarke, line 57: This tests that there were no exceptions thrown, but it would be good to test that the " Calif. Postcards Collection.  Collection 1351" string no longer has the space in front of it, too (I know it doesn't, but it's good to add that check to the test in case we ever break the functionality you've added). One way to do that would be to get `CsvMetadata metadata = myCsvParser.parse(getTestPath("spaces.csv")).getCsvMetadata()` and then check that the first string in the seventh string array (in the `getWorksList()` from `metadata`) starts with "C" instead of " ". Everything else looks good to me.

DRickard:
Tweaked the spaces test per Kevin's suggestion.

src/test/java/edu/ucla/library/iiif/fester/CsvParserTest.java
```diff
@@ -54,7 +55,9 @@ public final void testParse() throws CsvParsingException, CsvException, IOExcept
      */
     @Test
     public final void testParseSpaces() throws CsvParsingException, CsvException, IOException {
-        myCsvParser.parse(getTestPath("spaces.csv"));
+        final CsvMetadata metadata = myCsvParser.parse(getTestPath("spaces.csv")).getCsvMetadata();
+        final String first = metadata.getWorksList().get(8)[0];
```
ksclarke, line 59: Good, except it's the getWorksList().get(7)[0] record because the first non-header row in that CSV is a "collection" row (so doesn't show up in the `getWorksList()` result). You can confirm by commenting out your `trimValues(row);` code in the `CsvParser` and this test (with `metadata.getWorksList().get(8)[0]`) will still pass.

DRickard: Fixed the off-by-one issue in the test, and made resolution to do better job of learning data structure.

PR #124, opened 2020-07-14
[IIIF-897] Separate Festerize from Fester

markmatney:
Changes:
- move `src/main/scripts/festerize` to its own repository: https://github.com/UCLALibrary/festerize
- reference that repository in this project's README


ksclarke:
We also mention/link Festerize in https://github.com/UCLALibrary/fester/blob/IIIF-897/src/main/webroot/upload/csv/index.html#L93

markmatney:
thanks, ready for review again

PR #126, opened 2020-07-16
[IIIF-829] Check version of Festerize for compatibility

markmatney:
This implementation assumes that there is only ever a single version of Festerize that is compatible with a given version of Fester. To verify compatibility, the PostCsvHandler checks [the `User-Agent` header it receives with Festerize requests](https://github.com/UCLALibrary/festerize/pull/3) to see if it matches that single compatible version (which is now defined in the POM).


markmatney:
It would probably be good to add line endings to make the warning go away. Can't hurt. Hopefully we can set up our style checks to do this automatically at some point.

ksclarke:
Thanks for pointing that out, @DRickard ... I've created a ticket for it: https://github.com/ksclarke/freelib-build-tools/issues/12

PR #127, opened 2020-07-16
Iiif 881

DRickard:
Changes:

- Added select list for IIIF version number
- Made form fields required

### DIFF
--- a/src/main/webroot/upload/csv/index.html
+++ b/src/main/webroot/upload/csv/index.html
@@ -50,10 +50,10 @@
             </div>
             <div class="form-group">
               <label for="iiif-version">IIIF manifest version:</label> 
-	      <select name="iiif-version" id="iiif-version">
+	      <select name="iiif-version" id="iiif-version" required>
 		      <option value=""/>
-		      <option value="2">2.0</option>
-		      <option value="3">3.0</option>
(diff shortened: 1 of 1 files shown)


DRickard:
> Looks good, however IIRC we were planning to create a feature branch to work on exclusively for all these "Version Fester" tickets, so that we don't release a form to our end users that has an input which doesn't do anything. @ksclarke

That occurred to me during stand up; I suppose we could call this a code review, and then cancel the pull request until all the versioning updates are ready for release.

markmatney:
I was thinking that we'd create a new branch off of `main` for all the "Version Fester" tickets, which we'd submit PRs on, and then merge that branch into `main` when we're ready for that feature to go into production. That way, we don't have to wait for all those tickets to be done. @ksclarke what do you think?

ksclarke:
Yes, I think we could create a `version-fester` branch off of main and merge these PRs into that branch. When we've got them all ready to go we can merge that branch into main.

ksclarke:
I see one can edit a PR in the GitHub UI @DRickard so you could create `version-fester` branch and then adjust this PR to target that.

DRickard:
Created `version-fester` branch and pointed pull request to that.

PR #131, opened 2020-07-21
[Iiif 868] adding variable to OpenAPI spec

DRickard:
* Added the `iiif-version` variable to the `requestBody` element of `/collections`
* Used `enum` to specify legal values for new variable

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/Status.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/Status.java
@@ -12,6 +12,12 @@ public final class Status {
     /* Total memory property */
     public static final String TOTAL_MEMORY = "total_mem";
 
+    /* Used memory property */
+    public static final String USED_MEMORY = "used_mem";
+
+    /* Percentage memory ((used/total)*100) property */
+    public static final String PERCENT_MEMORY = "%_used_mem";
+
     /* Memory category for property values */
     public static final String MEMORY = "memory";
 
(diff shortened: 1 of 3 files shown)


src/main/webroot/upload/csv/index.html
```diff
@@ -40,14 +40,22 @@ <h1 id="csv-upload-title">CSV Upload</h1>
...
             <div class="form-group">
               <label for="iiif-host">IIIF Host:</label> <input type="text" value="https://iiif.library.ucla.edu/iiif/2"
-                name="iiif-host" id="iiif-host" onfocus="this.value=''" />
+                name="iiif-host" id="iiif-host" onfocus="this.value=''" required/>
               <div class="infotext">Use the default IIIF URL (i.e., https://iiif.library.ucla.edu/iiif/2) or enter the
                 base URL for a different IIIF server.</div>
             </div>
+            <div class="form-group">
+              <label for="iiif-version">IIIF manifest version:</label> 
+	      <select name="iiif-version" id="iiif-version" required>
```
ksclarke, line 53: Does the `required` do anything here? One of those three options is going to be submitted regardless.

I'm not sure what will happen when we load the OpenAPI spec and a blank value is submitted. The router may return an error since that value wouldn't conform to the spec? It will be something to test, I guess, once we hook that up.

It might also be interesting to get some user feedback (perhaps from Dawn and Lisa) about their preferences for a default value of 'v2' vs. the empty default value, if you haven't already. I'm not saying that needs to be done now... just that we might want to inquire before the publication of it (again, only if you didn't already). And if they did prefer a non-empty default value, at what point would we change that?

Anyway, looks good for now.

DRickard:
Before I squash and merge... I see I have the pull request set to merge into `main`; should I instead merge into `version-fester`?

PR #134, opened 2020-07-30
[IIIF-903] Update Fester dependencies

ksclarke:
Update two dependencies. The output can be checked by running this script locally on this branch: https://gist.github.com/ksclarke/18fa628d80a05411a4e6f958ba74cdc8

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,18 +69,21 @@
   </developers>
 
   <properties>
-    <vertx.version>3.9.1</vertx.version>
+    <vertx.version>3.9.2</vertx.version>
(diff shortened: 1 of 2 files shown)


ksclarke:
Hmm, I tried bumping up the Vert.x version (just a patch number increase) but that broke some things. I've rolled that version bump back, since it wasn't really related to this ticket, but I guess this is something we'll have to handle in the future. There weren't many breaking changes in the Vert.x release (https://github.com/vert-x3/wiki/wiki/3.9.2-Deprecations-and-breaking-changes) so it may be an unknown.

PR #135, opened 2020-08-03
[IIIF-915] Add Unit Test for CheckEndpointsHandler

DRickard:
Updates:
- Added CheckEndpointsHandlerTest.java
- Added debug message to fester_messages.xml

### DIFF
--- a/src/main/resources/fester_messages.xml
+++ b/src/main/resources/fester_messages.xml
@@ -161,7 +161,7 @@
   <entry key="MFS-145">S3 key already ends with .json extension: {}</entry>
   <entry key="MFS-146">Could not find {} manifest '{}' to update</entry>
   <entry key="MFS-147">Festerize is outdated, please upgrade to v{} (see README for instructions)</entry>
-  <entry key="MFS-148">S3 endpoints healthcheck failed: {})</entry>
-  <entry key="MFS-149">S3 endpoint response issue: {})</entry>
-  <entry key="MFS-150">Test CheckEndpointsHandler with request path: {})</entry>
+  <entry key="MFS-148">S3 endpoints healthcheck failed: {}</entry>
+  <entry key="MFS-149">S3 endpoint response issue: {}</entry>
+  <entry key="MFS-150">Test CheckEndpointsHandler with request path: {}</entry>
 </properties>
--- a/src/main/resources/fester_messages.xml
+++ b/src/main/resources/fester_messages.xml
@@ -163,4 +163,5 @@
   <entry key="MFS-147">Festerize is outdated, please upgrade to v{} (see README for instructions)</entry>
   <entry key="MFS-148">S3 endpoints healthcheck failed: {})</entry>
   <entry key="MFS-149">S3 endpoint response issue: {})</entry>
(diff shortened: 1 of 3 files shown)


src/test/java/edu/ucla/library/iiif/fester/handlers/CheckEndpointsHandlerTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+        webClient
+            .get(port, Constants.UNSPECIFIED_HOST, requestPath)
+            .expect(ResponsePredicate.SC_SUCCESS)
+            .as(BodyCodec.jsonObject())
+            .send(ar -> {
+                final int statusCode = ar.result().statusCode();
+
+                if (ar.succeeded()) {
+                    final JsonObject aResult = ar.result().body();
```
ksclarke, line 53: Oops, just noticed the aResult... we'd like to reserve the a* names for parameters that come in through a method so we can determine where they're defined.

src/test/java/edu/ucla/library/iiif/fester/handlers/CheckEndpointsHandlerTest.java
```diff
@@ -0,0 +1,66 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(CheckEndpointsHandlerTest.class, Constants.MESSAGES);
+
+    /**
+     * Test the CheckEndpointsHandler
+     *
+     * @param aContext A testing context
+     */
+    @Test
+    public void testCheckEndpointsHandler(final TestContext aContext) {
+        final String expectedStatus = Status.OK;
```
ksclarke, line 37: I guess also... what's the value to setting expectedStatus to the constant? Why not just use the constant where expectedStatus is used?

src/main/resources/fester_messages.xml
```diff
@@ -163,4 +163,5 @@
   <entry key="MFS-147">Festerize is outdated, please upgrade to v{} (see README for instructions)</entry>
   <entry key="MFS-148">S3 endpoints healthcheck failed: {})</entry>
   <entry key="MFS-149">S3 endpoint response issue: {})</entry>
+  <entry key="MFS-150">Test CheckEndpointsHandler with request path: {})</entry>
 </properties>
```
markmatney, line 167: these last few messages have unmatched closing parens

src/test/java/edu/ucla/library/iiif/fester/handlers/CheckEndpointsHandlerTest.java
DRickard: Changed variable name

src/test/java/edu/ucla/library/iiif/fester/handlers/CheckEndpointsHandlerTest.java
DRickard: Removed unneeded variable.

src/main/resources/fester_messages.xml
DRickard: Fixed the typos

PR #144, opened 2020-09-10
[IIIF-857] Version Fester

markmatney:
Enables the creation of manifests that conform to either v2 or v3 of the IIIF Presentation API.


markmatney:
The rewrites of ` src/main/java/edu/ucla/library/iiif/fester/handlers/CheckEndpointsHandler.java` and ` src/test/java/edu/ucla/library/iiif/fester/handlers/CheckEndpointsHandlerTest.java` are to normalize line endings to Unix-style, since those files were created on `main` after `version-fester` was created. I also removed an unnecessary try/catch construct in the former.

CODEOWNERS
```diff
@@ -2,4 +2,4 @@
 # This file contains a list of project codeowners
 #
 
-* @UCLALibrary/services-team-reviewers
```
ksclarke, line 5: Let's go ahead and leave this with services-team-reviewers... I thought that wasn't working well enough at one point and switched it back to individual names but have since concluded it's fine.

ksclarke (commented):
Could you make the one update to CODEOWNERS? It was my fault (but we might as well fix it now as we do the merge).

ksclarke (approved):

DRickard (approved):

PR #145, opened 2020-09-16
[IIIF-543] Travis build not downloading Fluency dependencies in container

cachemeoutside:


ksclarke (approved):

ksclarke:
This ticket says the dependencies aren't being downloaded. It's not clear to me why adding the profile to the install block would change this (they should be downloaded by the script block if they weren't downloaded in install), but I'm assuming you've tested this and it does work?

DRickard (approved):

PR #151, opened 2020-09-21
[IIIF-988] Fix thumbnail variable name

DRickard:
Updates: 
* Changed "THUMBNAIL" variable names and comments in Constants.java
* Updated V2ManifestVerticle.java to use new constant names
* Updated V3ManifestVerticle.java to use new constant names
* Updated MissingImageFT.java to remove "THUMBNAIL" names
FWIW: IIIF-988 references updating method names, but I didn't see any that used thumbnail or variations thereof.

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/Constants.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/Constants.java
@@ -239,14 +239,14 @@ public final class Constants {
     public static final String IIIF_API_V3 = "v3";
 
     /**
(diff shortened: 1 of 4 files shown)


ksclarke (dismissed):

markmatney (changes requested):
There are a couple variables with `THUMBNAIL` in their names in one of the test classes that I think should also be changed.

ksclarke (approved):

markmatney (approved):

PR #157, opened 2020-12-16
[IIIF-1029] Bump Festerize version

markmatney:
The next release of Festerize will be 0.2.0, which will add the capability of requesting updates to only the metadata of a manifest.


ksclarke (approved):

markmatney:
0.2.0 will also allow the user to choose which presentation API version to use when creating manifests. We had implemented this a while back but haven't put it out in a release yet.

DRickard (approved):

PR #160, opened 2021-01-27
[IIIF-1088] make v2 default option

DRickard:
Modified webroot/upload/csv/index.html to make "v2" option default selected, commented out empty select option.

### DIFF
--- a/src/main/webroot/upload/csv/index.html
+++ b/src/main/webroot/upload/csv/index.html
@@ -49,12 +49,11 @@
                 base URL for a different IIIF server.</div>
             </div>
             <div class="form-group">
-              <label for="iiif-version">IIIF manifest version:</label> 
-	          <select name="iiif-version" id="iiif-version" required>
-	              <!--option value=""/-->
-                  <option value="v2" selected>v2</option>
-                  <option value="v3">v3</option>
-              </select>
+              <label for="iiif-version">IIIF manifest version:</label><br/>
+	      <input type="radio" id="v2" name="iiif-version" value="v2" checked required>
+              <label for="v2">v2</label><br/>
+              <input type="radio" id="v3" name="iiif-version" value="v3">
+              <label for="v3">v3</label><br/>
             </div>
             <div class="form-group">
               <label for="metadata-update">Update work metadata only:</label> <input type="checkbox"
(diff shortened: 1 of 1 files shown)


src/main/webroot/upload/csv/index.html
```diff
@@ -51,8 +51,8 @@ <h1 id="csv-upload-title">CSV Upload</h1>
             <div class="form-group">
               <label for="iiif-version">IIIF manifest version:</label> 
 	          <select name="iiif-version" id="iiif-version" required>
-	              <option value=""/>
-                  <option value="v2">v2</option>
+	              <!--option value=""/-->
```
ksclarke, line 54: Why not delete instead of comment out?

DRickard: IIRC, the blank option was there to encourage user to think about what IIIF manifest version they wanted built; I'm guessing we'll want to go back to that approach once the v3 issue is solved.  If I'm wrong here, I'll delete the blank line.

ksclarke: Ah, okay, I see. I was thinking we'd want to make v3 the default once the client issue is resolved. @markmatney do you have opinions?

markmatney: I could see it either way. I do kind of like the user having to make a conscious choice, but a default value might reduce the likelihood of user error (where that default could perhaps be configured by an environment variable depending on the particular instance of Fester (each of which might have a different expected IIIF client) that the user is ingesting into).

FWIW, this could be implemented as two radio inputs where the user would still have to choose one in order to submit the form.

ksclarke: Geno said he could see v3 being the default (though didn't sound very opinionated about it). Waiting on Lisa's response.

I do like the idea of the radio buttons. They can start out unselected and then they give an either/or choice.

ksclarke: Lisa preferred a default value (v2 now and v3 in the future). I'd say let's remove the commented out blank value and give the radio buttons a try.

DRickard: Switched to radio buttons.

ksclarke (approved):

markmatney (approved):
I like the radio buttons

PR #161, opened 2021-01-29
[IIIF-1073] Add GitHub Actions

ksclarke:
* Update parent project reference
* Add GitHub Actions from parent project
* Add updated Checkstyle rules and config
* Update files to conform to Checkstyle rules
* Add uber-jar excludes to get rid of warning messages
* Update pom version to use the CI-friendly form

With this PR, I've also updated our issue/PR labels so that we can take advantage of the version incrementing in the Release Drafter Action. The Actions that have run as a part of this PR can be seen at https://github.com/UCLALibrary/fester/actions


DRickard (approved):
Won't claim this is a fully _informed_ review, but looks right to me.

.github/configs/release-drafter.yml
```diff
@@ -0,0 +1,41 @@
...
+  - title: '🌈 Documentation Updates'
+    labels:
+      - 'documentation'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'bugfix'
+  - title: '🧰 Breaking Changes'
+    label: 'breaking'
+version-resolver:
+  major:
```
markmatney, line 16: Are we going to start incrementing the major version number now?

ksclarke: I had this question too. One the one hand, we're using this in production in-house. On the other, would I consider it at a stable release if someone else other than us wanted to use it? I wanted to talk with you all about this to get thoughts on whether we should start using major versions for it.

markmatney: I guess I don't really have a strong opinion either way

markmatney (approved):

PR #162, opened 2021-02-03
[IIIF-1025] Endpoint to Add Thumbnail URLs

DRickard:
Updates;
* Created PostThumbnailsHandler.java for new endpoint
* Created ThumbnailUtils.java to support new endpoint
* Updated EndpointConfigHandler.java to add endpoint
* Updated Constants.java and Op.java with values related to new endpoint
* Updated fester.yaml to add endpoint specification
* Updated fester_messages.xml with endpoint-specific error message
* Created webroot/thumbnail/* for new upload page
* Created PostThumbFIT.java functional/integration test
* Created ThumbnailUtilsTest.java unit test
* Created allied.csv, empty_thumbs.csv, thumbs.csv for FIT
* Updated pom.xml, moving vertx-web-client into main project for use in PostThumbnailsHandler


ksclarke:
The codacy review has some good cleanup suggestions.

ksclarke:
To get rid of the build warning:
```
[INFO] --- maven-shade-plugin:2.4.3:shade (default) @ fester ---
[WARNING] vertx-web-client-3.9.4.jar, fester-0.0.0-SNAPSHOT.jar define 28 overlapping classes: 
[WARNING]   - io.vertx.ext.web.client.impl.predicate.ResponsePredicateResultImpl
[WARNING]   - io.vertx.ext.web.client.impl.HttpContext
[WARNING]   - io.vertx.ext.web.client.impl.predicate.PredicateInterceptor
[WARNING]   - io.vertx.ext.web.client.HttpResponse
[WARNING]   - io.vertx.ext.web.client.HttpRequest
[WARNING]   - io.vertx.ext.web.client.predicate.ResponsePredicateResult
[WARNING]   - io.vertx.ext.web.client.WebClientSession
[WARNING]   - io.vertx.ext.web.client.impl.HttpResponseImpl
[WARNING]   - io.vertx.ext.web.client.spi.CookieStore
[WARNING]   - io.vertx.ext.web.client.impl.WebClientInternal
[WARNING]   - 18 more...
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/
```
You can add `<exclude>io.vertx:vertx-web-client</exclude>` to the maven-shade-plugin's config in the POM.

src/main/webroot/thumbnail/css/upload-custom.css
```diff
@@ -0,0 +1,38 @@
...
+
+.infotext {
+  font-size: smaller;
+  color: #003B5C;
+}
+
+.form-group {
+  padding-top: 10px;
+}
+
```
markmatney, line 38: This file doesn't appear to be used -- delete?

src/main/java/edu/ucla/library/iiif/fester/utils/ThumbnailUtils.java
```diff
@@ -0,0 +1,80 @@
...
+   /**
+     * Adds a base IIF URL for the thumnail image.
+     *
+     * @param aCsvList A CSV in list of arrays format
+     * @param aURL The thumbnail URL
+   */
+    public static void addThumbnailURL(final List<String[]> aCsvList, final String aURL) {
+        Objects.requireNonNull(aCsvList);
+        final int thumbnailIndex = Arrays.asList(aCsvList.get(0)).indexOf(HEADER_THUMB);
+        for (int index = 1; index < aCsvList.size(); index++ ) {
```
markmatney, line 51: The control flow logic here and in the caller appears incorrect. The loop over the rows should be somewhere in the caller of this method, with each row getting a different thumbnail URL, instead of each row (work) being given the same thumbnail URL

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,207 @@
...
+            final FileUpload csvFile = csvUploads.iterator().next();
+            final String filePath = csvFile.uploadedFileName();
+            final String fileName = csvFile.fileName();
+
+            try {
+                final List<String[]> originalLines =
+                    new CSVReader(Files.newBufferedReader(Paths.get(filePath))).readAll();
+                final List<String[]> linesWithThumbs = ThumbnailUtils.addThumbnailColumn(originalLines);
+                final int manifestIndex = Arrays.asList(linesWithThumbs.get(0)).indexOf(CSV.MANIFEST_URL);
+                final String collectionURL = linesWithThumbs.get(1)[manifestIndex];
```
markmatney, line 115: The first line of a CSV is not necessarily a collection row (and I don't believe we are concerned with thumbnails of Collections, anyway) -- you can use `CsvParser.getObjectType` to determine the object type represented by the row

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,207 @@
...
+                final Future<JsonObject> collection = getManifest(collectionURL);
+                collection.onComplete(result -> {
+                    if (result.failed()) {
+                        returnError(response, HTTP.INTERNAL_SERVER_ERROR, result.cause().getMessage());
+                    } else {
+                        final JsonArray canvases = result.result().getJsonArray(MANIFESTS);
```
markmatney, line 121: This expression evaluates to an array of Manifests, not Canvases

(Even then it is specific to version 2.x of the IIIF Presentation API. Are we ready to support both API versions?)

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,207 @@
...
+                            canvasIndex = ThumbnailUtils.pickThumbnailIndex(canvasCount - 1);
+                        }
+                        final String canvasURL = canvases.getJsonObject(canvasIndex).getString(ID_TAG);
+                        final Future<JsonObject> canvas = getManifest(canvasURL);
+                        canvas.onComplete(canvasResult -> {
+                            if (canvasResult.failed()) {
+                                returnError(response, HTTP.INTERNAL_SERVER_ERROR, canvasResult.cause().getMessage());
+                            } else {
+                                final JsonObject thumbCanvas = canvasResult.result();
+                                final String thumbURL = thumbCanvas.getJsonArray("sequences").getJsonObject(0)
```
markmatney, line 136: (Just noting that this expression for getting the thumbnail URL is also v2-specific.)

markmatney (changes requested):

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
ksclarke, line 121: I'm a little confused by the variable name here. You're requesting the collection document (https://iiif.library.ucla.edu/collections/ark%3A%2F21198%2Fzz001ng4t6) and then getting the child work manifests of that collection from the returned JSON. I'm not sure why they're called canvases here.

If you pull each of the child manifests from your collection doc, you'll see the canvases in their JSON (e.g., https://iiif.library.ucla.edu/ark%3A%2F21198%2Fzz001p6m70/manifest). So, I think down below your `canvasURL` is really the `manifestURL`. You're then doing a `getManifest(URL)` on them and returning the work manifest.

Each of those work manifests (JSON) will have multiple canvases and I believe you'd want to pick the representative from those canvases for the work they're contained in (whereas now I think you'd random sampling the work manifests on the collection doc). You'd then need a way to associate the work you've pulled from the collection down with the work that's on the CSV and add its thumbnail base URL (it's ID, I'd assume).

If I'm misreading things, let me know. We can chat (or video) in Slack (which might be easier than commenting on here?)

ksclarke (changes requested):
I'm reading the ticket's "These should be base IIIF URLs, not full IIIF URLs to a thumbnail image (Hyrax will supply the full path for us)." as they should not have the full thumbnail URL with 600 size, but should just be the base IIIF URL, like: https://iiif.library.ucla.edu/iiif/2/ark%3A%2F21198%2Fzz001p6m70

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
ksclarke: Yes, good point. Another way to approach the problem would be to get the works from the CSV and query them for their canvases, skipping the collection altogether.

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
ksclarke: I think "yes" to "Are we ready to support both API versions."

markmatney: Some additional pointers that may be useful:

- To determine the API version, you can query the `@context` property of a Manifest or Collection
  - [v2](https://iiif.io/api/presentation/2.1/#linked-data-context-and-extensions)
  - [v3](https://iiif.io/api/presentation/3.0/#46-linked-data-context-and-extensions)
- For v2, the list of Canvases is at `manifest["sequences"][0]["canvases"]` (since all our v2 manifests should have only one Sequence)
- For v3, the list of Canvases is at `manifest["items"]`

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
markmatney: Just to elaborate a bit on [Kevin's comment about wanting to use base URLs](https://github.com/UCLALibrary/fester/pull/162#pullrequestreview-582618047), this is basically the URL of the IIIF Image API service for the image. To retrieve:
- v2: `canvas["images"][0]["resource"]["service"]["@id"]`
- v3: `canvas["items"][0]["items"][0]["body"]["service"]["id"]` (note no `@` for v3 IDs)

API version of the input data can be determined as explained [here](https://github.com/UCLALibrary/fester/pull/162#discussion_r569829714)

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
markmatney: agreed, best to just skip collection rows

DRickard:
Think I've fixed all issues and have made PR live again.

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,292 @@
...
+     *
+     * @param aColumnIndex The column in the CSV where thumbnail will be added
+     * @param aRowIndex The CSV row being updated
+     * @param aManifest A IIIF work manifest
+     * @param aCsvList A CSV file parsed as a list of string arrays
+     */
+    private void addV3Thumb(final int aColumnIndex, final int aRowIndex,
+                            final JsonObject aManifest, final List<String[]> aCsvList) {
+
+        final JsonArray canvases = aManifest.getJsonArray(ITEMS).getJsonObject(0).getJsonArray(ITEMS);
```
markmatney, line 220: For v3, the list of canvases is at `manifest["items"]`, not `manifest["items"][0]["items"]`.

Our manifests all (at the moment) have a 1:1 correspondence between canvases and images, so once you choose a  canvas with `canvasIndex` as you're already doing, you can get the image with `canvas["items"][0]["items"][0]["body"]`, then the IIIF service URL with `image["service"][0]["id"]`.

All these layers make a lot more sense in a linked data context (which were only using at a very surface level).

src/main/resources/fester.yaml
```diff
@@ -130,6 +130,58 @@ paths:
               schema:
                 type: string
                 format: binary
+  /thumbnails:
+    post:
+      tags: [Collection]
+      summary: Post a Collection to add thumbnails
+      description: "This endpoint accepts CSV data to use in the addition of a 
+        thumbnail to a collection"
```
markmatney, line 138: It would be more accurate to say that this is adding thumbnails to each work in a collection

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,292 @@
...
+            try (CSVReader csvReader = new CSVReader(Files.newBufferedReader(Paths.get(filePath)))) {
+                final CsvParser parser = new CsvParser().parse(Paths.get(filePath));
+                final List<String[]> originalLines = csvReader.readAll();
+                final List<String[]> linesWithThumbs = ThumbnailUtils.addThumbnailColumn(originalLines);
+                final int manifestIndex = Arrays.asList(linesWithThumbs.get(0)).indexOf(CSV.MANIFEST_URL);
+                final List<Future> futures = new ArrayList<>();
+                for (int rowIndex = 1; rowIndex < linesWithThumbs.size(); rowIndex++) {
+                    final ObjectType rowType = CsvParser.getObjectType(
+                                               linesWithThumbs.get(rowIndex), parser.getCsvHeaders());
+                    if (rowType.equals(ObjectType.WORK) || rowType.equals(ObjectType.PAGE)) {
```
markmatney, line 132: Are we doing this for pages as well as works? I think only works will have a manifest associated with them. @ksclarke ?

markmatney (changes requested):

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
ksclarke: Yes only works have manifests. Pages are all contained in works

src/test/resources/csv/hollywoodland.thumbs.csv
```diff
@@ -0,0 +1,10 @@
+"Project Name","Item ARK","Parent ARK","Item Status ID","Item Status","Object Type","File Name","Item Sequence","Duplicate","Visibility","Delete in Title","Type.typeOfResource","Subject","Type.genre","Format.medium","Alt ID.local","AltTitle.uniform","Title","Format.dimensions","Publisher.publisherName","Coverage.geographic","Format.extent","Relation.isPartOf","Description.note","Opac url","Finding Aid URL","Bucketeer State","IIIF Access URL","IIIF Manifest URL","thumbnail"
```
ksclarke, line 1: Noticing only `thumbnail` header is lower case. Perhaps we should upcase to make it consistent with the rest of the headers?

src/test/resources/csv/thumbs.csv
```diff
@@ -0,0 +1,6 @@
...
+Hathaway Manuscripts,ark:/21198/zz0009gv8j,ark:/21198/zz0009gsq9,Work,,1,text,Manuscripts,pd,"UCLA Charles E. Young Research Library Department of Special Collections, A1713 Young Research Library, Box 951575, Los Angeles, CA 90095-1575. E-mail: spec-coll@library.ucla.edu. Phone: (310)825-4989",lat,"University of California, Los Angeles. Library. Performing Arts Special Collections",Hathaway 1,170/768 v.1,"Breviary: Readings and Responsories.|~|Written in the Alemannic area, presumably in Switzerland.|~|Parchment (rather coarse), 13 long lines; ruling not discernible; pricking in the outer margin.  Written in a late, crude Caroline minuscule with squarish Alemannic features; dark brown ink.  Plain initials in red or stroked with red; initials in the text slashed with yellow; rubrics in red majuscule; neumes of St. Gall type.|~|Used for binding; folio 1 used as a pastedown: the verso is stained and rubbed.|~|ff. 1, 2 and 3 from the same manuscript.",Switzerland ?,202 x 135 mm,11th century,1001/1100,,,failed,,"https://iiif.library.ucla.edu/iiif/2/ark%3A%2F21198%2Fzz002czs4c/full/600,/0/default.jpg"
```
ksclarke, line 3: Why do the thumbnails in this file have full IIIF paths?

ksclarke (changes requested):
I notice that it looks like all our test works/CSVs have thumbnail URLs that are the same as their manifest URL. Is this a good test of the picking aspect or just the addition of the column? I think all of the samples only have one possible value? I did see there was a test for the index picker. It might be good to run that multiple times and confirm that you get different values (not just within the expected range, which is what's tested currently)? You'd have to account for the possibility that it's possible for it to return the same number twice, but that shouldn't happen multiple times in a row.

/me fights off urge to make obscure _Rosencrantz and Guildenstern Are Dead_ reference

src/test/java/edu/ucla/library/iiif/fester/fit/PostThumbFIT.java
```diff
@@ -0,0 +1,191 @@
...
+
+        /**
+         * Tests the V2 thumbnail workflow by posting a CSV.
+         *
+         * @param aContext A test context
+         */
+        @Test
+        public final void testV2ThumbCSV(final TestContext aContext) {
+            final Async asyncTask = aContext.async();
+            complete(asyncTask);
```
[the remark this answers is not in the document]
DRickard: Extraneous code... had disabled test cases while making large changes to PostThumbnailsHandler, forgot to remove that line when re-enabling.

src/test/java/edu/ucla/library/iiif/fester/fit/PostThumbFIT.java
```diff
@@ -0,0 +1,191 @@
...
+
+        /**
+         * Tests the V3 thumbnail workflow by posting a CSV.
+         *
+         * @param aContext A test context
+         */
+        @Test
+        public final void testV3ThumbCSV(final TestContext aContext) {
+            final Async asyncTask = aContext.async();
+            complete(asyncTask);
```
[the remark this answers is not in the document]
DRickard: Extraneous code... had disabled test cases while making large changes to PostThumbnailsHandler, forgot to remove that line when re-enabling.

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,292 @@
...
+            return ThumbnailUtils.pickThumbnailIndex(aCount - 1);
+        }
+    }
+
+    /**
+     * Return the processed CSV to user.
+     *
+     * @param aFileName Name of CSV file
+     * @param aFilePath Path to CSV file
+     * @param aCsvList CSV file paresed as list of string arrays
```
ksclarke, line 247: *parsed

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
DRickard: This is a V3 work manifest generated by TEST Fester: https://test-iiif.library.ucla.edu/ark%3A%2F21198%2Fzz001d2sw4/manifest; unless I've gone bonkers, it most definitely has `manifest["items"][0]["items"]`.

markmatney: It does, however, the value at `manifest["items"][0]["items"]` is a list of `AnnotationPage`s, not a list of `Canvas`es. For our items, at the moment there is only one AnnotationPage per canvas, but (potentially) multiple Canvases per manifest.

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,292 @@
+
+package edu.ucla.library.iiif.fester.handlers;
+
+import static info.freelibrary.iiif.presentation.v3.Constants.BODY;
+import static info.freelibrary.iiif.presentation.v3.Constants.CANVASES;
+import static info.freelibrary.iiif.presentation.v3.Constants.CONTEXT;
+import static info.freelibrary.iiif.presentation.v3.Constants.IMAGE_CONTENT;
+import static info.freelibrary.iiif.presentation.v3.Constants.RESOURCE;
+import static info.freelibrary.iiif.presentation.v3.Constants.SERVICE;
+import static info.freelibrary.iiif.presentation.v3.Constants.ITEMS;
```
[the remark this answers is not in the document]
DRickard: What's the Maven dependency for that version of JP?  Because the `jiiify-presentation-v2` currently in the project doesn't have a `Constants` class.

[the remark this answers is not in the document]
markmatney: For v2 the constants class is at `info.freelibrary.iiif.presentation.v2.utils.Constants`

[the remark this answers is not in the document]
ksclarke: It has it, but it's in the utils package: https://javadoc.io/doc/info.freelibrary/jiiify-presentation-v2/latest/info/freelibrary/iiif/presentation/v2/utils/Constants.html

I'm torn about where to put those Constants classes.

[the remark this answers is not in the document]
ksclarke: I do like the earlier idea of naming them something other than Constants. You'll get a class name conflict trying to use them both in the same class (one advantage of the v2/v3 verticles), but naming the class V2Constants or something would solve that issue.

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,292 @@
...
+
+/**
+ * Class to select and add thumbnail links to uploaded CSV
+*/
+public class PostThumbnailsHandler extends AbstractFesterHandler {
+
+    /* A logger for the class */
+    private static final Logger LOGGER = LoggerFactory.getLogger(PostThumbnailsHandler.class, Constants.MESSAGES);
+
+    private static final String SEQUENCES = "sequences";
```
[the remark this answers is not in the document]
DRickard: Switched to JIIFY Constants use

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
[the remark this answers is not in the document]
DRickard: Switched to using V2 Constants

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
DRickard: Modified logic

src/main/resources/fester.yaml
DRickard: Modified wording

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
DRickard: Removed PAGE clause

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,292 @@
...
+            final FileUpload csvFile = csvUploads.iterator().next();
+            final String filePath = csvFile.uploadedFileName();
+            final String fileName = csvFile.fileName();
+
+            try (CSVReader csvReader = new CSVReader(Files.newBufferedReader(Paths.get(filePath)))) {
+                final CsvParser parser = new CsvParser().parse(Paths.get(filePath));
+                final List<String[]> originalLines = csvReader.readAll();
+                final List<String[]> linesWithThumbs = ThumbnailUtils.addThumbnailColumn(originalLines);
+                final int manifestIndex = Arrays.asList(linesWithThumbs.get(0)).indexOf(CSV.MANIFEST_URL);
+                final List<Future> futures = new ArrayList<>();
```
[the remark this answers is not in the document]
DRickard: Added annotation

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,292 @@
...
+                promise.fail(asyncResult.result().statusMessage());
+            }
+        });
+
+        return promise.future();
+
+    }
+
+    /**
+     * Retireve the base thumbnail URL from IIIF V2 presentation manifests.
```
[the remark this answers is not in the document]
DRickard: Fixed

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
```diff
@@ -0,0 +1,292 @@
...
+        final JsonArray canvases = aManifest.getJsonArray(SEQUENCES).getJsonObject(0).getJsonArray(CANVASES);
+        final int canvasIndex = chooseThumbIndex(canvases.size());
+        final String thumbURL = canvases.getJsonObject(canvasIndex).getJsonArray(IMAGE_CONTENT)
+                                .getJsonObject(0).getJsonObject(RESOURCE).getJsonObject(SERVICE)
+                                .getString(Constants.ID_V2);
+        ThumbnailUtils.addThumbnailURL(aColumnIndex, aRowIndex, thumbURL, aCsvList);
+    }
+
+    /**
+     * Retireve the base thumbnail URL from IIIF V3 presentation manifests.
```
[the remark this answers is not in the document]
DRickard: Fixed

src/test/java/edu/ucla/library/iiif/fester/fit/PostThumbFIT.java
```diff
@@ -0,0 +1,191 @@
...
+
+                    if (statusCode == HTTP.OK) {
+                        final Buffer actual = response.body();
+                        final String contentType = response.getHeader(Constants.CONTENT_TYPE);
+                        final List<String[]> expected;
+
+                        try {
+                            expected = ThumbnailUtilsTest.read(V3_THUMB_CSV.getAbsolutePath());
+                            check(aContext, expected, actual);
+                        } catch (CsvException | IOException aDetails) {
```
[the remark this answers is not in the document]
DRickard: Changed variable name

src/test/resources/csv/hollywoodland.thumbs.csv
DRickard: Uppercased

src/test/resources/csv/thumbs.csv
DRickard: Forgot to update when switched to base URLs

src/main/java/edu/ucla/library/iiif/fester/handlers/PostThumbnailsHandler.java
DRickard: Dang it, Jim, I'm a coder, not a speller! /s
Fixed.

DRickard:
Did updates for all comments in last round, please re-re-review...

markmatney (approved):
:+1:

ksclarke (approved):

PR #164, opened 2021-02-10
[IIIF-1099] Add version check profile

ksclarke:
* Add a build profile (and script) for checking Docker image package versions
* Update obsolete image package versions and application dependency versions
* Changed 'install's in README to 'verify's (which makes more sense since we're not installing to a local Maven repo)
* Parameterized versions in Dockerfile
* Extended timeouts on some of the longer running tests
* Remove stray Travis references

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,8 +96,7 @@
     <!-- Versions of tools used for testing -->
     <jsoup.version>1.13.1</jsoup.version>
     <aws.sdk.version>1.11.950</aws.sdk.version>
-    <localstack.version>1.14.3</localstack.version>
-    <testcontainers.version>1.14.3</testcontainers.version>
+    <testcontainers.version>1.15.1</testcontainers.version>
     <jcip.annotations.version>1.0-1</jcip.annotations.version>
 
     <!-- Build plug-in versions -->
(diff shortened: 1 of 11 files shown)


ksclarke:
Had to update TestContainers to work around DockerHub's new rate limitations.

pom.xml
```diff
@@ -1183,6 +1019,263 @@
...
+                    <jdwp.host.port>${jdwp.host.port}</jdwp.host.port>
+                    <jdwp.client.config>${jdwp.client.config}</jdwp.client.config>
+                  </systemPropertyVariables>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
```
markmatney, line 1225: I'm curious why this change

markmatney (dismissed):
Just noticed a couple more `mvn install` references, and a question about the POM

DRickard (dismissed):

pom.xml
ksclarke: Moving this code into a default profile (that's activated by the absence of a `-Dversions` argument, ensures these plugins don't run when we're just running the versions profile (that profile just uses a standard ubuntu image so the tests would fail against it).

DRickard (approved):

pom.xml
markmatney: ah I see, that makes a lot of sense

markmatney (approved):

PR #171, opened 2021-03-24
[IIIF-1129] Enable creation of v3 manifests with video content 

markmatney:
Changes (6e3a652):
- make `CsvParser` aware of A/V headers
- disallow submission of CSVs that contain A/V headers when IIIF Presentation API 2 is requested (`CsvParser.checkIiifPresApiCompatibility`)
- move `AbstractFesterVerticle.getMetadata` to `CsvParser`, and add an additional `getMetadata` implementation that does some basic type validation @ L472 (which I am interested in feedback on)
- make `V3ManifestVerticle` aware of video


markmatney:
Oh, and `src/test/resources/csv/video/video.csv` can be used to create a manifest that renders successfully in a locally-hosted UV instance.
image

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
```diff
@@ -304,6 +328,65 @@ private void extractPageMetadata(final String... aRow) throws CsvParsingExceptio
...
+     * Checks for A/V metadata in rows that represent v2 canvases, and returns the row if none is found.
+     *
+     * @param aRow A row from the metadata CSV
+     * @param aPath A path to a CSV file
+     * @return The row
+     * @throws CsvParsingException If the row represents a v2 canvas and contains any A/V metadata
+     * @throws MimeTypeParseException
+     */
+    @SuppressWarnings("unchecked")
+    private String[] checkIiifPresApiCompatibility(final String[] aRow, final Path aPath, final String aIiifVersion)
```
ksclarke, line 341: A very Java method name ;-)

I wonder whether `checkApiCompatibility` would have been enough, given the context of the application? No problem if the answer is "no" though.

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
```diff
@@ -304,6 +328,65 @@ private void extractPageMetadata(final String... aRow) throws CsvParsingExceptio
...
+            throws CsvParsingException {
+        final String rowId = getMetadata(aRow, myCsvHeaders.getItemArkIndex()).get();
+
+        final Optional<Integer> mediaWidth =
```
ksclarke, line 345: Heh, was just re-debating the internal use of Optional(s) again with myself with the ticket I'm working on. They do make using things that are... well, optional... easier.

src/main/java/edu/ucla/library/iiif/fester/verticles/V3ManifestVerticle.java
```diff
@@ -414,61 +417,82 @@ private void updatePages(final Message<JsonObject> aMessage) throws JsonProcessi
...
-                        } else {
-                            size = width;
-                        }
+            // We've already validated the MIME type in CsvParser, so it's fine to just check for a substring here
+            if (format.isPresent() && format.get().contains("video/")) {
+                final VideoContent video;
+
+                resourceURI = CsvParser.getMetadata(columns, aCsvHeaders.getAudioVideoAccessUrlIndex()).get();
+                video = new VideoContent(resourceURI);
+                // TODO: set format on video
```
ksclarke, line 437: Was leaving this TODO in intentional?

ksclarke (dismissed):
Looks good. Nice commenting throughout.

I do have a question about the TODO (should it have been implemented)?

ksclarke:
Codacy does have some things to say about the PR. Some are good. Others you can ignore by using Checkstyle/PMD SuppressWarnings annotations.

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
markmatney: agreed, will change

src/main/java/edu/ucla/library/iiif/fester/verticles/V3ManifestVerticle.java
markmatney: I am actually not sure. Currently, `format` is set automatically if the resource id has a filename extension that JP can infer a MimeType from (also, the `setFormat` method is not visible on `VideoContent`). I forgot to remove this comment when I realized this, but now I realize that this behavior might not be desirable (e.g. if the id ends in `.m3u8`).

DRickard (dismissed):

src/main/java/edu/ucla/library/iiif/fester/verticles/V3ManifestVerticle.java
ksclarke: Do you think `setFormat()` should be visible, overriding the default detection?

markmatney: I think it would be good to have the option to use it, in case the URL has no extension or if the extension is incorrect.

ksclarke: Would you create a ticket on that project. I don't think it affects us for our first round of materials though. We're controlling the extension (mp4).

ksclarke:
Since we discussed the apps teams not seeing UV working with some media, I'm also curious what version of local UV did you test with for this?

markmatney:
Tested with UV 3.1.1.

src/main/java/edu/ucla/library/iiif/fester/verticles/V3ManifestVerticle.java
markmatney: Okay, great. And [done](https://github.com/ksclarke/jiiify-presentation/issues/94).

markmatney:
Justifications for the PMD suppressions:
- the long variable name is for a private member variable in what I'd classify as boilerplate (so I wouldn't expect its name to be read too often by human eyes), it's not causing any line wrapping, and changing the name would break the symmetry between member variable name and corresponding getter/setter which I think should be preserved
- the two exceptions that don't preserve the stack trace aren't hiding anything that I'd consider useful for debugging
- there will eventually be another branch on the switch statement to handle audio

ksclarke (approved):
👍🏼

DRickard (approved):

PR #172, opened 2021-03-25
[IIIF-1136] Update manifest labels on metadata update requests

markmatney:
Also, `src/test/resources/csv/hathaway-updated.csv` now differs from `src/test/resources/csv/hathaway.csv` by only the single line that we test (L3).


src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
```diff
@@ -326,6 +326,12 @@ private void updateWork(final Message<JsonObject> aMessage) throws JsonProcessin
         final DeliveryOptions options = new DeliveryOptions();
         final JsonObject message = new JsonObject();
 
+        CsvParser.getMetadata(workRow, csvHeaders.getTitleIndex()).ifPresentOrElse(title -> {
```
ksclarke, line 329: I had a question about this when Dawn mentioned she wanted _all_ metadata updated. Are we now updating all the metadata that we use in the manifest?

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
```diff
@@ -357,23 +357,23 @@ private void updateWorks(final CsvHeaders aCsvHeaders, final CsvMetadata aCsvMet
         final Promise<LockedManifest> promise = Promise.promise();
         final String collectionID = aCsvMetadata.getFirstCollectionID(aCsvHeaders.getParentArkIndex()).get();
 
-        // If we were able to get a lock on the manifest, update it with our new works
+        // If we were able to get a lock on the collection, update it with our new works
         promise.future().onComplete(handler -> {
             if (handler.succeeded()) {
                 final Map<String, List<String[]>> worksMap = aCsvMetadata.getWorksMap();
-                final LockedManifest lockedManifest = handler.result();
+                final LockedManifest lockedCollection = handler.result();
```
ksclarke, line 364: No objections to this, but isn't it kind of a superficial change if we keep `LockedManifest` and `getLockedManifest()`? The distinction we're making is in variable name only. I'm wondering if you'd see this change as a stepping stone to something more comprehensive(?)

ksclarke (dismissed):

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
markmatney: Maybe this method should be renamed as well? `getLockedIiifResource`? In general I'm interested in making the manifest/collection distinction more clear throughout the codebase.

markmatney: And LockedManifest to LockedIiifResource?

src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
markmatney: I'm not sure, although I think it might be good to double check with Dawn that our implementation is updating the rest of the metadata as desired.

markmatney: (But I'd vote for that being separate from this ticket.)

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
ksclarke: Makes sense to me 👍🏼

ksclarke: I do wonder whether the `Iiif` part is necessary... is LockedResource too generic (even in the context of the app)?

src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
ksclarke: Maybe mention in standup today if she's there? I'm for us implementing what the ticket says, but perhaps surface the question in case she had something different in mind than what she wrote in the ticket.

markmatney: :+1: will do

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
markmatney: I don't think it makes a huge difference, but I do think the `Iiif` part helps me more easily remember what the method/class does/is when I come across it in the code

markmatney:
argh, I should have put the file renaming and the content changes in two commits

ksclarke (approved):

DRickard (approved):

PR #176, opened 2021-04-08
[IIIF-1143] Update A/V URL Code

DRickard:
Updates:
* Removed references to "A/V Access URL"
* Renamed `myAudioVideoAccessUrlIndex` (and related methods) to `myContentAccessUrlIndex`
* Updated `CsvParser` to distinguish between image and A/V URLs in IIIF Access URL field
* Fixed some missing @Test entries in `PostCsvFIT`
* Updated A/V test resources

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,7 +152,7 @@
     <fester.http.port>8888</fester.http.port>
 
     <!-- String that should be present in A/V access URLs -->
-    <fester.av.string>pairtree</fester.av.string>
+    <fester.av.string>https://wowza.library.ucla.edu/iiif_av_public/</fester.av.string>
 
     <!-- The version of Festerize that is compatible with this version of Fester -->
     <festerize.version>0.2.0</festerize.version>
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,9 @@
     <!-- The port Fester listens on -->
     <fester.http.port>8888</fester.http.port>
 
+    <!-- String that should be present in A/V access URLs -->
+    <fester.av.string>pairtree</fester.av.string>
+
     <!-- The version of Festerize that is compatible with this version of Fester -->
     <festerize.version>0.2.0</festerize.version>
 
--- a/src/main/docker/configs/fester.properties.tmpl
+++ b/src/main/docker/configs/fester.properties.tmpl
@@ -24,7 +24,7 @@ fester.http.port=$FESTER_HTTP_PORT
 iiif.base.url=$IIIF_BASE_URL
 
 # String that should be present in A/V access URLs
-fester.av.string=AV_URL_STRING
+fester.av.string=$AV_URL_STRING
 
 # The version of Festerize that is compatible with this version of Fester
 festerize.version=$FESTERIZE_VERSION
--- a/src/main/docker/configs/fester.properties.tmpl
+++ b/src/main/docker/configs/fester.properties.tmpl
@@ -23,5 +23,8 @@ fester.http.port=$FESTER_HTTP_PORT
 # The base URL for the IIIF image service (includes service prefix)
 iiif.base.url=$IIIF_BASE_URL
 
+# String that should be present in A/V access URLs
+fester.av.string=AV_URL_STRING
+
 # The version of Festerize that is compatible with this version of Fester
 festerize.version=$FESTERIZE_VERSION
--- a/src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
@@ -41,6 +41,8 @@ public class CsvParser {
 
     private static final Pattern EOL_PATTERN = Pattern.compile(".*\\R");
 
+    private static final String AV_URL_STRING = "pairtree";
+
     private final Map<String, List<String[]>> myWorksMap = new HashMap<>();
 
     private final Map<String, List<String[]>> myPagesMap = new LinkedHashMap<>();
@@ -337,6 +339,7 @@ public class CsvParser {
      * @throws CsvParsingException If the row represents a v2 canvas and contains any A/V metadata
      */
     @SuppressWarnings("unchecked")
+    @SuppressWarnings("BooleanExpressionComplexity")
     private String[] checkApiCompatibility(final String[] aRow, final Path aPath, final String aIiifVersion)
             throws CsvParsingException {
         final String rowId = getMetadata(aRow, myCsvHeaders.getItemArkIndex()).get();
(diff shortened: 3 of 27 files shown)


src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
```diff
@@ -348,11 +350,13 @@ private void extractPageMetadata(final String... aRow) throws CsvParsingExceptio
...
 
         if (Constants.IIIF_API_V2.equals(aIiifVersion)) {
+            //if (hasAVContent(mediaWidth, mediaHeight, mediaDuration, mediaFormat, audioVideoAccessUrl)) {
             if (mediaWidth.isPresent() || mediaHeight.isPresent() || mediaDuration.isPresent() ||
-                    mediaFormat.isPresent() || audioVideoAccessUrl.isPresent()) {
+                    mediaFormat.isPresent() || (audioVideoAccessUrl.isPresent() &&
+                    audioVideoAccessUrl.toString().contains(AV_URL_STRING))) {
                 throw new CsvParsingException(MessageCodes.MFS_168, rowId, aPath);
             }
         } else { // Constants.IIIF_API_V3
```
markmatney, line 362: In the case of v3, might we also want to check if the access URL _doesn't_ contain `AV_URL_STRING` (and throw the exception if so) in the switch statement below?

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
```diff
@@ -41,6 +41,8 @@
 
     private static final Pattern EOL_PATTERN = Pattern.compile(".*\\R");
 
+    private static final String AV_URL_STRING = "pairtree";
```
markmatney, line 44: Where does this value come from? https://docs.library.ucla.edu/pages/viewpage.action?pageId=237784030#DLCSA/VAssetDocumentation-StructureoftheStreamingDirectories (`pairtree_root`)?

markmatney (commented):
Looks good, a few questions though

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
```diff
@@ -348,11 +350,13 @@ private void extractPageMetadata(final String... aRow) throws CsvParsingExceptio
         final Optional<Float> mediaDuration =
                 (Optional<Float>) getMetadata(aRow, myCsvHeaders.getMediaDurationIndex(), Float.class, aPath);
         final Optional<String> mediaFormat = getMetadata(aRow, myCsvHeaders.getMediaFormatIndex());
-        final Optional<String> audioVideoAccessUrl = getMetadata(aRow, myCsvHeaders.getAudioVideoAccessUrlIndex());
+        final Optional<String> audioVideoAccessUrl = getMetadata(aRow, myCsvHeaders.getContentAccessUrlIndex());
 
         if (Constants.IIIF_API_V2.equals(aIiifVersion)) {
+            //if (hasAVContent(mediaWidth, mediaHeight, mediaDuration, mediaFormat, audioVideoAccessUrl)) {
```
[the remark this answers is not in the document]
DRickard: Nerts--thought I removed all the commented-out code.

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
ksclarke: Hard coding this value to something in our AV URL path prevents others from possibly using the application. I'd suggest making the A/V server base URL configurable and comparing that configured value to the found URL.

ksclarke (changes requested):

src/main/docker/configs/fester.properties.tmpl
```diff
@@ -23,5 +23,8 @@ fester.http.port=$FESTER_HTTP_PORT
 # The base URL for the IIIF image service (includes service prefix)
 iiif.base.url=$IIIF_BASE_URL
 
+# String that should be present in A/V access URLs
+fester.av.string=AV_URL_STRING
```
ksclarke, line 27: Needs a $ in front of it.

pom.xml
```diff
@@ -151,6 +151,9 @@
     <!-- The port Fester listens on -->
     <fester.http.port>8888</fester.http.port>
 
+    <!-- String that should be present in A/V access URLs -->
+    <fester.av.string>pairtree</fester.av.string>
```
ksclarke, line 155: While technically this works fine, I think it would be more illustrative of the type of value someone else would put in here to use the base URL for our Wowza server instead of just "pairtree", which probably won't hold much significance to someone else (other than us) reading the code.

src/main/java/edu/ucla/library/iiif/fester/verticles/ManifestVerticle.java
```diff
@@ -103,7 +104,8 @@ public void start(final Promise<Void> aPromise) {
                 final String action = message.headers().get(Constants.ACTION);
                 final Path filePath = Paths.get(body.getString(Constants.CSV_FILE_PATH));
                 final String iiifVersion = body.getString(Constants.IIIF_API_VERSION);
-                final CsvParser csvParser = new CsvParser().parse(filePath, iiifVersion);
+                final CsvParser csvParser = new CsvParser().setAVUrlString(config().getString(Config.AV_URL_STRING)) //, DEFAULT_AV_STRING))
```
markmatney, line 107: Since this line is calling the constructor, why not pass the configuration directly to it, rather than creating a new public setter method? If the programmer forgets the call to setAVUrlString, the code will throw a NullPointerException from within the `parse` method, which could potentially be confusing (admittedly not much of a practical concern now, since this is the only place its used, but I think in general we should try not to write code that allows errors like that)

ksclarke (dismissed):

markmatney (approved):

markmatney (changes requested):
I take that approval back -- looks like the build is failing. Fails in about 10 sec when I run `mvn clean verify` locally

ksclarke (dismissed):
👍🏼

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
```diff
@@ -503,4 +524,15 @@ public static ObjectType getObjectType(final String[] aRow, final CsvHeaders aCs
...
+    /**
+     * Sets the A/V URL identifier string.
+     *
+     * @param aAVUrlString The A/V URL identifier string
+     * @return This CSV parser
+     */
+    public CsvParser setAVUrlString(final String aAVUrlString) {
+        myAVUrlString = aAVUrlString;
+        return this;
+    }
```
markmatney, line 537: unused method should be removed

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
```diff
@@ -49,13 +49,25 @@
...
+    public CsvParser(final String aAVUrlString) {
+        myAVUrlString = aAVUrlString;
+    }
 
+    /**
+     * Creates a new CsvParser.
+     *
+     */
+    public CsvParser() {
+        myAVUrlString = Constants.DEFAULT_AV_STRING;
```
markmatney, line 70: I think this constructor should also be removed since it isn't being used

markmatney (changes requested):

src/main/java/edu/ucla/library/iiif/fester/CsvParser.java
ksclarke: I could see the argument for keeping this and the setter instead of the constructor that takes the av string argument. I think the original concern was there was a possibility of a null pointer in case someone used the constructor but didn't set the av string. Here there is a default value set that one can override by using the setter.

ksclarke: To make it more generally useful, one might make the default value `http://` or something that would match on everything. In that case, you'd override by using the setter to set your specific A/V server.

markmatney: FTR, in a separate channel we decided to revert to the no-arg constructor and instead pass the A/V URL prefix string to `parse` (reverting my earlier suggestion -- sorry!) which sets us up nicely for rewriting it as a static method

markmatney (approved):

ksclarke (approved):

markmatney (approved):

PR #177, opened 2021-04-09
Update MIME type database so Fester maps .mpd URLs to application/dash+xml

markmatney:
Changes:
- add a custom `content-types.properties` file that adds the MPEG DASH (`application/dash+xml`) type to the OpenJDK 11 MIME type database
  - the existing database was retrieved with the Eclipse debugger; I set a breakpoint in the `sun.net.www.MimeTable` class where it loads the database into a Properties object, which I then serialized, and put into the aforementioned file along with a new entry for the MPEG DASH type
- add a test to confirm that Fester recognizes `.mpd` URLs as MPEG DASH

One thing I'm not sure about is if `src/main/docker/configs` is the right place for this file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -533,6 +533,7 @@
           <runArgs>
             <runArg>-Dvertx-config-path=${project.basedir}/target/test-classes/test-config.properties</runArg>
             <runArg>-Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory</runArg>
+            <runArg>-Dcontent.types.user.table=src/main/docker/configs/content-types.properties</runArg>
           </runArgs>
         </configuration>
       </plugin>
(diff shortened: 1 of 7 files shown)


ksclarke (dismissed):
It needs to go where you put it because the Docker builder can't reach outside of the `src/main/docker` hierarchy. And, it is a configuration of the container.

Am I correct in understanding that, though you're configuring it for the vert.x maven plugin, there is nothing that's currently using/testing that configuration? The functional test is testing against the application running in the container. Did you just add the other for symmetry or because you expect people to need to test it outside of the container?

markmatney:
Okay, thanks.

Within the `runDebug` profile it is necessary/used, but I'm not sure about the other one.

ksclarke (approved):
👍🏼

DRickard (approved):

PR #182, opened 2021-04-27
Add support for Choice in A/V streams

ksclarke:
* Eclipse is doing reformatting/cleanup (new version of Eclipse; I probably need to resync the formatting rules)
* Update jiiify-presentation-v3 (which meant a few method name changes, e.g.: `addCanvases()` instead of `addCanvas()`)
* Add content type for vnd.apple.mpegURL
* Add configurable extensions so MPD and HLS streams are used when creating a Choice
* Changed DEFAULT_AV_STRING since MPD is no longer the default; using "mp4" instead
* Updated tests to merge in the configuration values from the test config file

A/V Pairtree needs to be updated before deploying these changes (here is that related PR https://github.com/UCLALibrary/av-pairtree/pull/3). I also confirmed the A/V Pairtree changes didn't break anything with services-metagetter.


DRickard (dismissed):

src/main/resources/fester_messages.xml
```diff
@@ -189,4 +189,5 @@
   <entry key="MFS-176">Found item '{}' with format '{}' in CSV '{}', but the access url has not been processed for Wowza</entry>
   <entry key="MFS-177">Test server was responsive: {}</entry>
   <entry key="MFS-178">Looking for test application's server port: {}</entry>
+  <entry key="MFS-179">A/V server URL substitution only supports up to one replacement: {}</entry>
```
markmatney, line 192: This makes me curious as to why `WatcherVerticle.addIdPath` in av-pairtree may return a URL template string with two substitution patterns.

src/test/resources/test-config.properties
```diff
@@ -25,5 +25,9 @@ iiif.base.url=${iiif.base.url}
 # String that should be present in A/V access URLs
 fester.av.string=${fester.av.string}
 
+# The extension(s) to be appended onto A/V server URLs for a choice of streams; a comma should be used as the substring
+# delimiter if there is more than one extension (e.g., /manifest.mpd|/playlist.m3u8)
```
markmatney, line 29: Should a comma be the delimiter here instead of a `|`?

markmatney (commented):
A couple questions (one related to av-pairtree, which I'd like to understand before reviewing that PR)

src/test/resources/test-config.properties
ksclarke: Yes, I changed and missed updating this comment. Will fix.

src/main/resources/fester_messages.xml
ksclarke: I started with supplying the `mp4:` part too as a part of the substitutions. I decided that wasn't really necessary for Fester, but didn't see a reason to remove that additional functionality from av-pairtree. Someone might find being able to do substitutions on either side of the ID/Pairtree path useful, depending on their image server.

markmatney: :+1:

markmatney:
Looks like there's a new JRE available -- `11.0.11+9-0ubuntu2~20.04`

markmatney (approved):

DRickard (approved):

PR #186, opened 2021-04-30
Store our A/V thumbnail URLs in Docker env instead of hard-coding

markmatney:
I figured I'd make these changes we've discussed today while I was thinking about it. The behavior is the same as before, except these changes would allow someone using this app to omit thumbnails from A/V canvases, if desired, by removing the URLs from the Docker config.

### DIFF
--- a/src/main/docker/configs/fester.properties.default
+++ b/src/main/docker/configs/fester.properties.default
@@ -32,10 +32,10 @@ AV_URL_STRING=https://wowza.library.ucla.edu/iiif_av_public/
 AV_EXTS=/manifest.mpd,/playlist.m3u8
 
 # The URL of the default thumbnail image for Canvases with Sound content
-AV_DEFAULT_AUDIO_THUMBNAIL_URL=https://prod-uclalibrary-resources.s3-us-west-2.amazonaws.com/audio_icon.svg
+AV_DEFAULT_AUDIO_THUMBNAIL_URL=
 
 # The URL of the default thumbnail image for Canvases with Video content
-AV_DEFAULT_VIDEO_THUMBNAIL_URL=https://prod-uclalibrary-resources.s3-us-west-2.amazonaws.com/video_icon.svg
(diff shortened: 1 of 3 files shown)


ksclarke:
Hmm, I was thinking that we'd still want to have them in the manifests (for when our manifests are harvested) but that the CSS would just override the default thumbnail value on the front-end site. Nothing wrong with remove our defaults in the app (since others might want to use it without our defaults), but you're planning after this to set those config values to our defaults or ??

markmatney:
This PR does set those config values to our defaults, so running the app as-is would still put thumbnails on A/V canvases

ksclarke:
Ah, I missed that you were already doing what I was suggesting: `AV_DEFAULT_AUDIO_THUMBNAIL_URL=https://prod-uclalibrary-resources.s3-us-west-2.amazonaws.com/audio_icon.svg` Nevermind. :-)

ksclarke (dismissed):
👍🏼

DRickard (dismissed):

markmatney:
I think it makes the code easier to understand as well as more generic

ksclarke:
I think I'd prefer it in the helm charts and just leave it out of the docker config, but it's not a big deal to me. I've already approved the PR. 😄

markmatney:
If that's preferable then let's do it; for me, the part about being "more generic" is more about those strings living in the environment configuration instead of in the Java code

ksclarke:
I agree they're better in environment configs since they're not generic images. Up to you. I just said I thought adding them in the helm charts was preferable, but I'm not dictating a path here (don't have that strong of feelings about it).

ksclarke (approved):

markmatney:
Oh, I see why they were in the default config -- the tests need those values

ksclarke:
You can supply them via the test ENV in the surefire config, fwiw

markmatney:
ah right -- D'oh!

DRickard (approved):

ksclarke:
@markmatney Can this be closed now?

PR #193, opened 2021-07-20
[IIIF-1059] Update status handler with " MB"

angelahuqing:
GetStatusHandler
converted long memory values into strings with " MB" attached to string
changed memory.put to send new string values
GetStatusFT
retrieved string values and used lambda function to split string values where there is a space
Check if second word after space in string is "MB"
*Note
Ultimately did not use SizeFromBytes in FileUtils because whenever I would use it, GetStatusFT would fail (sometimes it would be GB instead of MB and there isn't a way, at least to my knowledge, to specify MB)

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/handlers/GetStatusHandler.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/handlers/GetStatusHandler.java
@@ -38,6 +38,10 @@ public class GetStatusHandler implements Handler<RoutingContext> {
             final long usedMem = totalMem - freeMem;
             final double percentMem = (double) usedMem / (double) totalMem * 100D;
             final JsonObject memory = new JsonObject();
+            final String totalMemStr = totalMem + Constants.SPACE + Constants.MB_STR;
+            final String freeMemStr = freeMem + Constants.SPACE + Constants.MB_STR;
+            final String usedMemStr = usedMem + Constants.SPACE + Constants.MB_STR;
+
 
             if (percentMem >= WARN_PERCENT && percentMem < ERROR_PERCENT) {
                 status.put(Status.STATUS, Status.WARN);
@@ -47,8 +51,8 @@ public class GetStatusHandler implements Handler<RoutingContext> {
                 status.put(Status.STATUS, Status.OK);
             }
             status.put(Status.MEMORY, memory);
-            memory.put(Status.TOTAL_MEMORY, totalMem).put(Status.FREE_MEMORY, freeMem).put(Status.USED_MEMORY, usedMem)
-                    .put(Status.PERCENT_MEMORY, percentMem);
+            memory.put(Status.TOTAL_MEMORY, totalMemStr).put(Status.FREE_MEMORY, freeMemStr)
+                .put(Status.USED_MEMORY, usedMemStr).put(Status.PERCENT_MEMORY, percentMem);
 
             response.setStatusCode(200);
             response.putHeader(Constants.CONTENT_TYPE, Constants.JSON_MEDIA_TYPE).end(status.encodePrettily());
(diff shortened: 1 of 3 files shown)


ksclarke:
You're correct that there isn't a way to specify MB with SizeFromBytes. It just uses the highest denomination it can. Having it come out with GB instead of MB for this ticket would have been fine -- the point is just to have an indication of what sort of number we're looking at, but not using it is fine too.

src/main/java/edu/ucla/library/iiif/fester/handlers/GetStatusHandler.java
```diff
@@ -38,6 +40,9 @@ public void handle(final RoutingContext aContext) {
             final long usedMem = totalMem - freeMem;
             final double percentMem = (double) usedMem / (double) totalMem * 100D;
             final JsonObject memory = new JsonObject();
+            final String totalMemStr = totalMem + MB_STR;
```
DRickard, line 43: Perhaps use `concat()` for these three vars; or `StringBuffer`/`StringBuilder`, as those are more efficient in building strings from components?

src/test/java/edu/ucla/library/iiif/fester/fit/GetStatusFT.java
```diff
@@ -47,4 +51,9 @@ public final void testStatus(final TestContext aContext) {
         });
     }
 
+    @FunctionalInterface
```
DRickard, line 54: What's the reason for using an interface, instead of just doing a direct `String.split()`?

angelahuqing: I think it's because the code checks would fail whenever I did String.spilt(" ") because " " was repeated three times. I could have just put " " in a variable instead, but I had never written an interface so I wanted to try that. If you think it's better though I can change it to use String.split()

DRickard: FWIW, I think the constant and `String.split()` would be simpler to maintain... but perhaps Mark or Kevin will convince me otherwise.

src/main/java/edu/ucla/library/iiif/fester/handlers/GetStatusHandler.java
ksclarke: This isn't really necessary anymore. Newer versions of the JDK use StringBuilder under the hood in the JDK to optimize the concatenation.

src/test/java/edu/ucla/library/iiif/fester/fit/GetStatusFT.java
ksclarke: The checkstyle rule complains about the multiple " "s because they are allocating multiple strings for the same value. We have a `Constants.EMPTY` variable defined for " " so you can just use that. You can also use a static import to get that down to just `EMPTY` in the code.

DRickard: On reread; two points: 1) since you're using MB_STR in both handler and test, should probably be added to Constants.java, for consistency/simplciity; 2) you probably don't need the split code--just do a `String.contains()` for the constant.

angelahuqing: Sounds good. Is there somewhere specific I should add it in the file or do I just add it to the end? I'll change it to String.contains as well. That makes a lot more sense.

ksclarke: Oops, my mistake... that should have been Constants.SPACE, but that's only defined in the newer version of freelib-utils. For now, it's probably easier to just put a new SPACE variable (" ") in Fester's Constants class.

ksclarke (dismissed):
👍🏼

ksclarke (dismissed):
👍🏼

src/test/java/edu/ucla/library/iiif/fester/fit/GetStatusFT.java
angelahuqing: I think I finished fixing everything so that it follows all of your suggestions.

DRickard (dismissed):

DRickard (dismissed):

DRickard (approved):

markmatney (approved):
👍🏻 , good pull request title as well

ksclarke (approved):
👍🏼

PR #196, opened 2021-08-18
[IIIF-492] Refactored ordinary async.complete() calls to more cautious calls

angelahuqing:
- Added 
 if (!async.isCompleted())
     { async.complete(); }
as a replacement to async.complete() in various places


src/test/java/edu/ucla/library/iiif/fester/handlers/DeleteManifestHandlerTest.java
```diff
@@ -44,7 +44,9 @@ public void testDeleteManifestHandler(final TestContext aContext) throws IOExcep
 
             if (response.statusCode() == HTTP.SUCCESS_NO_CONTENT) {
                 aContext.assertFalse(myS3Client.doesObjectExist(myS3Bucket, myManifestS3Key));
-                asyncTask.complete();
+                if (!asyncTask.isCompleted()) {
```
DRickard, line 47: Why not just call `complete(asyncTask)`, as you do in `PutManifestHandlerTest`?

DRickard (commented):
Since you have an identical `complete(Async)` method defined in multiple classes, perhaps you should pull that out as a `static` method in a utility class?

src/test/java/edu/ucla/library/iiif/fester/handlers/DeleteManifestHandlerTest.java
angelahuqing: I should call complete(asyncTask). I think I missed that because previously in this same file  
if (!asyncTask.isCompleted()) ... was placed in it. I'll change both instances of it.

markmatney (dismissed):
There are still a few more of these to change. You can use something like this to find them all (or most of them anyway; we tend to name these variables `asyncTask` or `aAsyncTask` or `myAsyncTask`):
```bash
git grep -C 10 "\(a\|my\)\?[Aa]syncTask\.complete()"
```
*edit*: a better command

angelahuqing:
> There are still a few more of these to change. You can use something like this to find them all (or most of them anyway; we tend to name these `asyncTask` or `aAsyncTask`):
> 
> ```shell
> git grep -C 10 "Task\.complete"
> ```
> 
> _edit_: a better command

Thanks for letting me know. The only other instance that I see is myAsyncTask.complete() in ServerCheck.java. Did you catch any other ones besides that?

angelahuqing:
> Since you have an identical `complete(Async)` method defined in multiple classes, perhaps you should pull that out as a `static` method in a utility class?

Where would you suggest placing the utility class?

markmatney (commented):
Here's where I'm seeing them (line numbers found using the `-n` flag of `git grep`):

src/test/java/edu/ucla/library/iiif/fester/fit/MissingImageFT.java:180
src/test/java/edu/ucla/library/iiif/fester/fit/PagesManifestFT.java:179
src/test/java/edu/ucla/library/iiif/fester/handlers/CheckEndpointsHandlerTest.java:54
src/test/java/edu/ucla/library/iiif/fester/handlers/GetCollectionHandlerTest.java:65
src/test/java/edu/ucla/library/iiif/fester/handlers/GetCollectionHandlerTest.java:96
src/test/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandlerTest.java:106
src/test/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandlerTest.java:148
src/test/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandlerTest.java:195
src/test/java/edu/ucla/library/iiif/fester/utils/ServerChecker.java:64

markmatney:
Ah, yes, you are correct about the one with myAsyncTask inside the loop

DRickard:
> > Since you have an identical `complete(Async)` method defined in multiple classes, perhaps you should pull that out as a `static` method in a utility class?
> 
> Where would you suggest placing the utility class?

In the TEST `edu.ucla.library.iiif.fester.utils` package, possibly as a new method in the `TestUtils` class.

angelahuqing:
> Here's where I'm seeing them (line numbers found using the `-n` flag of `git grep`):
> 
> src/test/java/edu/ucla/library/iiif/fester/fit/MissingImageFT.java:180
> src/test/java/edu/ucla/library/iiif/fester/fit/PagesManifestFT.java:179
> src/test/java/edu/ucla/library/iiif/fester/handlers/CheckEndpointsHandlerTest.java:54
> src/test/java/edu/ucla/library/iiif/fester/handlers/GetCollectionHandlerTest.java:65
> src/test/java/edu/ucla/library/iiif/fester/handlers/GetCollectionHandlerTest.java:96
> src/test/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandlerTest.java:106
> src/test/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandlerTest.java:148
> src/test/java/edu/ucla/library/iiif/fester/handlers/PutCollectionHandlerTest.java:195
> src/test/java/edu/ucla/library/iiif/fester/utils/ServerChecker.java:64

Ahh I get it now. Those were all files that had previously replaced async.complete() with the more cautious call, but they aren't using the method complete(asyncTask) used in other files. Since I'm going to create a static method in a utility class now, I'll change all of those.

angelahuqing:
> > > Since you have an identical `complete(Async)` method defined in multiple classes, perhaps you should pull that out as a `static` method in a utility class?
> > 
> > 
> > Where would you suggest placing the utility class?
> 
> In the TEST `edu.ucla.library.iiif.fester.utils` package, possibly as a new method in the `TestUtils` class.

I'll start working on that now. Thanks for the suggestion.

markmatney:
@angelahuqing that is a good point. This ticket has been in the backlog for a couple years now and has gone somewhat out of date. I have updated it to be more specific (basically just incorporating the feedback we've given so far). Thank you!

ksclarke (approved):
👍🏼 Looks good. The TestUtils.complete() is a much nicer way to handle the case across Verticle(s) and Handler(s) than we previously had.

DRickard (approved):

PR #198, opened 2021-09-13
[IIIF-998] Add thumbnails to v3 Canvases that have Image resources

markmatney:


src/main/java/edu/ucla/library/iiif/fester/Constants.java
```diff
@@ -279,6 +279,18 @@
...
+     * The string template of default thumbnail URIs.
+     */
+    @SuppressWarnings({ "PMD.LongVariable" })
+    public static final String IIIF_THUMBNAIL_URI_TEMPLATE = "{}/full/{}/0/default.jpg";
+
+    /**
+     * The default size of thumbnail images.
+     */
+    @SuppressWarnings({ "PMD.LongVariable" })
+    public static final String DEFAULT_IMAGE_THUMBNAIL_SIZE = "!200,200";
```
DRickard, line 292: I'm guessing the "!" in the value is a typo...

markmatney: This is actually part of the Image API syntax; it's basically saying "hey image server, give me an image that fits within a bounding box 200px wide and 200px high". See https://iiif.io/api/image/2.1/#size

ksclarke (approved):

DRickard (approved):

PR #202, opened 2022-01-05
[SERV-313] Update dependencies for security

ksclarke:
* Fix multiple security issues with dependencies
* Change one test fixture to remove the ".0" decimal on serialization of duration float value
* Modify PostCsvFIT test with a code optimization Eclipse automatically made
* Add String -> JsonObject conversions now that the upstream jpv3 doesn't output JsonObject(s)

### DIFF
--- a/src/test/java/edu/ucla/library/iiif/fester/utils/TestUtils.java
+++ b/src/test/java/edu/ucla/library/iiif/fester/utils/TestUtils.java
@@ -1,7 +1,12 @@
 
 package edu.ucla.library.iiif.fester.utils;
 
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import info.freelibrary.util.StringUtils;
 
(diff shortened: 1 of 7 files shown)


src/test/java/edu/ucla/library/iiif/fester/utils/TestUtils.java
```diff
@@ -44,79 +56,32 @@ public static String getTestData(final Vertx aVertx, final String aPath) {
...
-
-        return distinctManifestCount == 1;
+    private static JsonObject sanitize(final JsonObject aManifest, final String[] aKeyArray) {
+        String json = aManifest.encode();
+
+        for (final String name : aKeyArray) {
+            json = json.replaceAll(StringUtils.format(ID_PATTERN, name), StringUtils.format(EMPTY_ID, name));
+        }
+
+        return new JsonObject(json);
```
markmatney, line 84: I appreciate that this is a much simpler implementation, but I'm not sure that we don't need to check _any_ ids. Should we test that ids of A/V resources end with `/manifest.mpd` or `/playlist.m3u8`, for example?

ksclarke: The tests where we're using this don't seem to be relevant to this. I could see an argument for testing those, but those should be more specific tests in my opinion rather than structural tests that also look at ID values.

markmatney: Not relevant even though 3 of the 4 tests we're using this on involve A/V material?

ksclarke: Yes, if we want to validate A/V materials' IDs, we should have specific tests for that. Keeping the tests focused on one thing makes them more useful and easier to maintain.

ksclarke: Resolution: we chatted more in Slack. I'm going to implement a check that includes the non-random parts of the IDs too here.

markmatney:
FWIW, if you apply this patch:

```patch
diff --git a/src/test/resources/json/v3/video_mpd.json b/src/test/resources/json/v3/video_mpd.json
index bbc5e72..22059c5 100644
--- a/src/test/resources/json/v3/video_mpd.json
+++ b/src/test/resources/json/v3/video_mpd.json
@@ -18,7 +18,7 @@
       "id": "http://b1dbe4a0-443c-479f-bf0a-25c352df0d8f.iiif.library.ucla.edu/ark%3A%2F21198%2Fzz00000001/manifest/canvas-s39o",
       "width": 720,
       "height": 540,
-      "duration": 5021.0,
+      "duration": 5021,
       "thumbnail" : [
         {
           "type" : "Image",
```

the failing test can be made to pass without rewriting the helper methods.

I guess JP was serializing whole-number durations as floats before?

markmatney (approved):

PR #227, opened 2022-10-13
Update dependencies

ksclarke:
Some dependencies (security and base image) need updating. I see the last GHA CI build failed because of this. Still looking into the S3 verticle not being found issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
       MAVEN_CACHE_KEY: ${{ secrets.MAVEN_CACHE_KEY }}
     strategy:
       matrix:
-        java: [ 11, 17 ]
+        java: [ 17 ]
 
     steps:
       - name: Check out code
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       MAVEN_CACHE_KEY: ${{ secrets.MAVEN_CACHE_KEY }}
     strategy:
       matrix:
-        java: [ 11 ]
+        java: [ 11, 17 ]
```
markmatney, line 17: I am wondering whether it would be good to stick to a single Java version. If we specify more than one then that means we are restricted to the language features of the earliest one. I personally am interested in utilizing new language features.

DRickard (approved):

.github/workflows/build.yml
ksclarke: So, you're proposing updating to 17 when we touch things (and can do so without complications)?

markmatney: yes

ksclarke: I'm good with this, too, if we land on the latest LTS as the release version. We can add newer JDKs as alternative test versions.

markmatney: Ah, I like that

PR #228, opened 2022-11-04
[SERV-588] Remove failed Vert.x upgrade [dcf32e2]

ksclarke:
This is a patch for a rollback + cherry-picking of later commits.


ksclarke:
Ignoring Codacy for now. The base project was updated (so stricter Checkstyle/PMD rules), and I made all the adjustments in the commit that's being rolled back now. For now, I'm just ignoring the stylistic changes that need to be made. I will do a second commit with them (to keep it separate from the functional changes in this PR).

PR #230, opened 2023-09-12
[SERV-910] Add a catch for duplicate IDs in collection doc

ksclarke:
* Catch unusual duplicate ID runtime exception

And, in addition:
* Switch to our new convention for listing devs
* Update dependencies
* Eclipse did two minor syntactical reformats


ksclarke:
Abandoned PR

PR #238, opened 2024-03-20
[SERV-909] Update metadata updating

ksclarke:
Previously, if the manifest didn't have any metadata we didn't update it, but this commit changes that so that new metadata fields are added even if there aren't any pre-existing metadata fields in the record.

I also turned down logging on the presentation library's attempt to derive a media type from the URL since not all URLs have file extensions on them. This is something that should probably be fixed upstream too, fwiw.

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
@@ -610,6 +610,9 @@ public class V2ManifestVerticle extends AbstractFesterVerticle {
                 } // If label matches but we have no updated value, ignore the metadata
             }
         } else {
+            // Our metadata comes in pairs so, when looping through an array of all the values, we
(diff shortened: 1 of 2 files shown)


src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
```diff
@@ -609,6 +609,10 @@ private Metadata updateMetadata(final Metadata aMetadata, final String... aStrin
                     metadata.add(entry);
                 } // If label matches but we have no updated value, ignore the metadata
             }
+        } else {
+            for (int index = 0; index < aStringArray.length / 2; index += 2) {
```
DRickard, line 613: Not saying this is wrong, but would like an explanation for the `aStringArray.length / 2` and i`ndex += 2` values in the loop.

DRickard (approved):

angelahuqing (approved):

PR #239, opened 2024-04-05
[SERV-1068] Only set fallback thumbail on static images

ksclarke:
If we set thumbnails for images served by IIIF we'd have to supply the IIIF thumbnail dims, like we do in the v3 manifests. We've decided for now (cf ticket) to just not set those thumbnails, since we were not doing that before and the IIIF viewers know how to handle creating thumbnails when a specific one isn't included. This leaves the thumbnail to be set in v2 manifests only for the static images, where we have w/h supplied to the process via the CSV.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -15,3 +15,7 @@ ignore:
     - '*':
         reason: Service is deprecated; moving to GoLang
         expires: 2024-12-01
+  'SNYK-JAVA-IONETTY-6483812':
+    - '*':
+        reason: Service is deprecated; moving to GoLang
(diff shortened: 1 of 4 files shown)


ksclarke:
Converted to draft since tests failed. I forgot I'd probably need to unwind some of those changes too.

ksclarke:
Now container packages need updating too.

angelahuqing (dismissed):

angelahuqing (approved):

DRickard (approved):

PR #242, opened 2025-04-02
[SERV-1249] Fix GA YAML syntax

ksclarke:
* Fix GA YAML syntax issue (not currently using a linter, but perhaps we should be)


DRickard (approved):

ksclarke:
Thanks for the review, @DRickard, but I somehow messed up this PR. I've opened a new/correct one at https://github.com/UCLALibrary/fester/pull/243

PR #244, opened 2025-07-17
[SERV-1299] Fix manifest PUT

ksclarke:
* Update the manifest PUT endpoint to use the S3BucketVerticle instead of the S3 client
* Eclipse did some auto-formatting
* Update AbstractFesterHandlerTest's S3 configuration so that it uses the more recent method
* Update manifest PUT's tests to also use the more recent method


ksclarke:
Converted to draft because the Docker image needs some updates, too, it seems.

DRickard (approved):

PR #270, opened 2026-05-01
[DLSR-317] Improve bad URL handling

ksclarke:
Have the ImageInfoLookup throw an ImageNotFoundException if the image couldn't be found due to a bad/empty URL. Catch this and return a 400 (Bad Request) to fail the job.

* Turn down log levels for easier, everyday log surfing
* Fail job if the job has been submitted with a bad/empty image URL in the CSV data

### DIFF
--- a/src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
+++ b/src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
@@ -406,6 +406,7 @@ public class V2ManifestVerticle extends AbstractFesterVerticle {
      *
      * @param aMessage A message with information about the page updates
      * @throws JsonProcessingException If there is trouble deserializing message components
+     * @throws ImageNotFoundException If the canvas' image couldn't be found
      */
     private void updatePages(final Message<JsonObject> aMessage)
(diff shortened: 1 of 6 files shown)


src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
```diff
@@ -400,7 +407,8 @@ private void updateWork(final Message<JsonObject> aMessage) throws JsonProcessin
      * @param aMessage A message with information about the page updates
      * @throws JsonProcessingException If there is trouble deserializing message components
      */
-    private void updatePages(final Message<JsonObject> aMessage) throws JsonProcessingException {
+    private void updatePages(final Message<JsonObject> aMessage)
```
DRickard, line 411: I think you need to add the `ImageNotFoundException` to the `@throws` javadoc line

DRickard (dismissed):

src/main/java/edu/ucla/library/iiif/fester/verticles/V2ManifestVerticle.java
ksclarke: Thanks! Will do.

DRickard (approved):

